PR #395, opened 2022-05-05
Simplify Volume phase

adracus:
'Roll back' similar to old behavior: `Phase` field exists again, but
additionally we also maintain a timestamp for its last transition time.
Makes code simpler + status easier to read.


afritzler (approved):

adracus:
Merging as the check passed locally + we're anyways switching runners.

PR #421, opened 2022-05-17
Update prerequisites documentation in local `kind` setup 

lukasfrank:
# Proposed Changes
- Updated prerequisites in for local kind setup in docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,9 @@ APISERVER_IMG ?= apiserver:latest
 # Docker image name for the mkdocs based local development setup
 IMAGE=onmetal-api/documentation
 
+# kind cluster which should be used
+KIND_CLUSTER ?= main
+
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
 GOBIN=$(shell go env GOPATH)/bin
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -6,6 +6,9 @@ APISERVER_IMG ?= apiserver:latest
 # Docker image name for the mkdocs based local development setup
 IMAGE=onmetal-api/documentation
 
+# kind cluster which should be used
+KIND_CLUSTER ?= kind
```
afritzler, line 10: Instead changing the `Makefile` I would suggest to put this into the local setup guide and mention there that the current kind `context` should be set to `kind`.

afritzler (changes requested):
Can you also give this PR a shorter and more meaningful title + a few words in the PR description about the change itself.

Makefile
lukasfrank: Sure that might be an option. however if someone created a kind cluster with a name different from the defaults it will not work anymore. The consequence is that you need to parse through the entire Makefile and find the places where the kind cluster is referenced. 

If you guys never change the kind content I totally agree that the setup guidelines are enough.

afritzler: The problem is, that we can not - and should not - assume how the users local `kind` setup looks like. That is why we should stick to the default case as much as possible in the beginning.

afritzler:
Can you also please adjust the PR title?

Makefile
lukasfrank: Got it.

docs/development/setup.md
```diff
@@ -34,6 +42,9 @@ the manifests is available via
 make kind-install
 ```
 
+**Note**: If a `kind context` different to the default is used, be aware that the `Makefile` is referencing the default
```
afritzler, line 45: I would note here, that the user should make sure the `kind get clusters` is pointing to a `kind` cluster in case he has multiple environments running.

afritzler (changes requested):

afritzler (approved):

PR #462, opened 2022-08-03
Add `NetworkProtection` controller

### LINKED ISSUE #444: Finalizer for the Network object
# Summary

It would be nice to have finalizers also on the Network object, while there are NetworkInterface/s referring to the specific Network object. This should be useful while NetworkInterface objects are being deleted/reconciled, they might need to access certain info from the Network object.

This has a stronger use case at the Virtualization layer, as the Network object there has an ID in it.

afritzler:
# Proposed Changes

The `NetworkProtection` controller ensures that a finalizer is present on the `Network` resource as long as it is being referenced/used by resources like `NetworkInterface`s or `AliasPrefix`es.

Fixes #444

### DIFF
--- a/controllers/networking/network_protection_controller.go
+++ b/controllers/networking/network_protection_controller.go
@@ -1,0 +1,187 @@
+/*
+ * Copyright (c) 2022 by the OnMetal authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package networking
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"github.com/onmetal/controller-utils/clientutils"
+	networkingv1alpha1 "github.com/onmetal/onmetal-api/apis/networking/v1alpha1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	networkFinalizer     = "networking.api.onmetal.de/network"
+	networkNameFieldPath = ".spec.networkRef.name"
+)
+
+type NetworkProtectionReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+//+kubebuilder:rbac:groups=networking.api.onmetal.de,resources=networks,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=networking.api.onmetal.de,resources=networks/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=networking.api.onmetal.de,resources=networks/finalizers,verbs=update
+//+kubebuilder:rbac:groups=networking.api.onmetal.de,resources=networkinterfaces,verbs=get;list;watch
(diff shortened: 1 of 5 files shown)


controllers/networking/network_protection_controller.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	return r.reconcile(ctx, log, network)
+}
+
+func (r *NetworkProtectionReconciler) delete(ctx context.Context, log logr.Logger, network *networkingv1alpha1.Network) (ctrl.Result, error) {
+	log.Info("Deleting Network")
+
+	if networkInUse, err := r.isNetworkInUse(ctx, log, network); err != nil {
+		return ctrl.Result{}, err
+	} else if !networkInUse {
```
adracus, line 73: To avoid `if / else` here, I'd recommend
```go
if ok, err := r.isNetworkInUse(ctx, log, network); err != nil || !ok {
  return ctrl.Result{}, err
}
```
this should handle both cases, avoid the `else` and makes the indentation depth less.

controllers/networking/network_protection_controller.go
```diff
@@ -0,0 +1,187 @@
...
+	return ctrl.Result{}, nil
+}
+
+func (r *NetworkProtectionReconciler) isNetworkInUse(ctx context.Context, log logr.Logger, network *networkingv1alpha1.Network) (bool, error) {
+	log.V(1).Info("Checking if the Network is in use")
+
+	// NetworkInterfaces
+	networkInterfaces := &networkingv1alpha1.NetworkInterfaceList{}
+	if err := r.List(ctx, networkInterfaces, &client.ListOptions{
+		FieldSelector: fields.OneTermEqualSelector(networkNameFieldPath, network.Name),
```
adracus, line 115: Use `client.MatchingFields` as a shorthand for this. Same for `aliasPrefixes`.

controllers/networking/network_protection_controller.go
```diff
@@ -0,0 +1,187 @@
...
+		if _, err := clientutils.PatchEnsureNoFinalizer(ctx, r.Client, network, networkFinalizer); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to remove finalizer from network: %w", err)
+		}
+	}
+
+	log.Info("Successfully deleted Network")
+	return ctrl.Result{}, nil
+}
+
+func (r *NetworkProtectionReconciler) reconcile(ctx context.Context, log logr.Logger, network *networkingv1alpha1.Network) (ctrl.Result, error) {
```
adracus, line 84: How does Kubernetes do the finalizer management in case of the `pv-protection` controller? Don't they always add the finalizer and only decide when to remove it? Would simplify the logic a bit more.

adracus (commented):
Please double-check with the kubernetes pv-protection controller how caching might affect the removal of the finalizer - do they ever do a live check (w/o cache) agains the kube-apiserver?

afritzler:
Thanks @adracus for the review! Regarding the non cached access: from looking at the [pv protection controller]( https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/volume/pvprotection/pv_protection_controller.go#L127) there is only one `GET` for the PV object via the `PersistentVolumeLister`. Determining if a PV is being used is done just by checking the `Phase` of the PV - so pretty trivial to what we need to do here.

controllers/networking/network_protection_controller.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	if ok, err := r.isNetworkInUse(ctx, log, network); err != nil || ok {
+		return ctrl.Result{}, err
+	} else {
+		log.V(1).Info("Removing finalizer from Network as the Network is not in use")
+		if _, err := clientutils.PatchEnsureNoFinalizer(ctx, r.Client, network, networkFinalizer); err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to remove finalizer from network: %w", err)
+		}
+	}
```
adracus, line 98: Remove this part and only use it during `delete`.

controllers/networking/network_protection_controller.go
```diff
@@ -0,0 +1,178 @@
...
+
+	log.Info("Successfully deleted Network")
+	return ctrl.Result{}, nil
+}
+
+func (r *NetworkProtectionReconciler) reconcile(ctx context.Context, log logr.Logger, network *networkingv1alpha1.Network) (ctrl.Result, error) {
+	log.Info("Reconcile Network")
+
+	log.V(1).Info("Patching finalizer from Network as the Network is in use")
```
adracus, line 86: Log message might be confusing, should be `Ensuring finalizer`

adracus (changes requested):

adracus (approved):

PR #504, opened 2022-10-04
Common `ResourceName`s for `VolumeClass`

lukasfrank:
# Proposed Changes

- Define `ResourceName`s for `VolumeClass` that storage providers can use constants to check for default resources

### DIFF
--- a/apis/storage/volumeclass_types.go
+++ b/apis/storage/volumeclass_types.go
@@ -21,6 +21,15 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// Max throughput in bytes per seconds.
+	ResourceBPS corev1.ResourceName = "bps"
+	// Max IOPS in input/output operations per second.
+	ResourceIOPS corev1.ResourceName = "iops"
+	// Dynamic resource limits flag: limits  per GB of volume.
+	ResourceDynamic corev1.ResourceName = "dynamic"
+)
+
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +genclient
 // +genclient:nonNamespaced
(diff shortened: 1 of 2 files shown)


apis/storage/volumeclass_types.go
```diff
@@ -21,6 +21,15 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	// Max throughput in bytes per seconds.
+	ResourceBPS corev1.ResourceName = "bps"
```
adracus, line 26: we should calls this `tps` - the value would then end up as
```yaml
tps: 1Gi
```
which is more in line with e.g. Kubernetes' `memory`

apis/storage/volumeclass_types.go
```diff
@@ -21,6 +21,15 @@ import (
...
+	// Max IOPS in input/output operations per second.
+	ResourceIOPS corev1.ResourceName = "iops"
+	// Dynamic resource limits flag: limits  per GB of volume.
+	ResourceDynamic corev1.ResourceName = "dynamic"
```
adracus, line 30: we should call this `maxSize`

adracus (changes requested):
We're missing these values in the versioned types.
Also, the Godoc of this should be of the form
`// <name> <explanation>`

apis/storage/volumeclass_types.go
lukasfrank: With `ResourceDynamic` I wanted to express that the limits of the `VolumeClass` are absolute limits or 'dynamic' limits. 
In case if the `ResourceDynamic` key is present the limits are multiplied with the requested size of the `Volume`. 

It might be better to put such a flag  in the spec of the `VolumeClass` itself.

apis/storage/volumeclass_types.go
lukasfrank: Sure. What does `tps` stand for?

adracus (commented):
Please still fix the godoc + add the values to the versioned types.

adracus (approved):

PR #518, opened 2022-10-18
Loadbalancer and NAT-Gateway Proposal

gehoern:
# Proposed Changes

Create a proposal for the `LoadBalancer` and `NATGateway` types.
These types are essential for exposing services to the internet / share internet access over multiple `NetworkInterface`s.


docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+    - [Details](#details)
+- [Proposal](#proposal)
+
+## Summary
+Network service for Load Balancing is needed in every modern infrastructure (e.g. Kubernetes). Load balancing makes services scalable, fault-tolerant and accessible from external users. The load balancer described here is a L3-Loadbalancer (so it implements an IP level load balancing). The load balancer function is a standard network function that can be to be deployed (as an implementation detail) on a platform/node that is capable of running network functions (e.g. dp-service enabled nodes/nic's)
```
adracus, line 33: We should omit the part about how load balancers might be deployed, especially mentioning `dp-service`.

docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+    - [Details](#details)
+- [Proposal](#proposal)
+
+## Summary
+Network service for Load Balancing is needed in every modern infrastructure (e.g. Kubernetes). Load balancing makes services scalable, fault-tolerant and accessible from external users. The load balancer described here is a L3-Loadbalancer (so it implements an IP level load balancing). The load balancer function is a standard network function that can be to be deployed (as an implementation detail) on a platform/node that is capable of running network functions (e.g. dp-service enabled nodes/nic's) 
+
+## Motivation
+The onMetal network is a fully routed network. To run workloads hosted on several nodes (in Kubernetes terms equivalent to cluster IP's) and not only on a single node (node IP's) solutions like ECMP (Equal cost multi pathing) seem to be the natural solution. But ECMP has certain limitations: the amount of nodes it can be used to distribute traffic to and a hash based algorithm to pin the network flows, that changes - based on the amount of targets - its distributed pinning.
```
adracus, line 36: Please replace `onMetal` with `onmetal` (applies for all occurrences)

adracus, line 36: The motivation here goes into the direction of how an LB can be implemented. In `onmetal-api` we should rather focus on the general motivation why Load Balancers are necessary. In our case e.g., we have the `VirtualIP` which is a single stable public IP attached to a `NetworkInterface`, however, to scale beyond single network interfaces, `LoadBalancer`s offer the capability to dynamically route traffic to multiple targets.

docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Motivation
+The onMetal network is a fully routed network. To run workloads hosted on several nodes (in Kubernetes terms equivalent to cluster IP's) and not only on a single node (node IP's) solutions like ECMP (Equal cost multi pathing) seem to be the natural solution. But ECMP has certain limitations: the amount of nodes it can be used to distribute traffic to and a hash based algorithm to pin the network flows, that changes - based on the amount of targets - its distributed pinning. 
+To avoid those limitations and to have better control over the load balancer behavior an onMetal load balancer service needs to be provided.
+
+### Goals
+- define a user facing API (onMetal api) for the load balancer
+- enable the user to request an L3 Loadbalancer (only IP addresses) with load balancer address (outside facing) and
```
adracus, line 41: Meld into sth like `Define an API for managing L3 load balancers with publicly available addresses`

docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Motivation
+The onMetal network is a fully routed network. To run workloads hosted on several nodes (in Kubernetes terms equivalent to cluster IP's) and not only on a single node (node IP's) solutions like ECMP (Equal cost multi pathing) seem to be the natural solution. But ECMP has certain limitations: the amount of nodes it can be used to distribute traffic to and a hash based algorithm to pin the network flows, that changes - based on the amount of targets - its distributed pinning. 
+To avoid those limitations and to have better control over the load balancer behavior an onMetal load balancer service needs to be provided.
+
+### Goals
+- define a user facing API (onMetal api) for the load balancer
+- enable the user to request an L3 Loadbalancer (only IP addresses) with load balancer address (outside facing) and 
+- with multiple target's based on the onMetal ([OEP-1](01-networking-integration.md)) network interface architecture
+- the targets of the load balancer must be dynamic changeable
```
adracus, line 43:
```suggestion
- the targets of the load balancer should be dynamically changeable
```

docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+- the load balancer needs to avoid forwarding unintended traffic.
+  - based on ports/protocols (UDP/TCP/SCTP) it must filter what needs to be load balanced
+  - ICMP requests are not part of forwarded traffic and will be filtered by the load balancer anyhow (the loa)
```
adracus, line 46: What about

> (the loa)

?

docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+- Load balancer (IP) and targets can be parts of different networks but
+- all targets (NetworkInterfaces) must be of the same network
```
adracus, line 48:
```suggestion
- all targets (`NetworkInterface`s) must be in the same `Network` 
```

docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Non-Goals
+- no address or port translation / rewriting (no SNAT / DNAT) (L4 Loadbalancer) support
+- no injection of additional information (e.g. x-forwarded-for) (L7 Loadbalancer) support
+- no protocol offloading like ssl (L7 Loadbalancer) support
+- if multiple load balancer IP's are needed, request multiple load balancers (a load balancer services one load balancer IP)
+
+### Details
+- load balancing is used to deliver a packet addressed to the load balancer to one of its targets via the onMetal network routing
+- the load balancing will be for the target and the source transparent
```
adracus, line 58:
```suggestion
- the load balancing will transparent for both the target and the source
```

docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+- the target needs to know the load balancers IP and needs to answer with it (and to receive traffic with it)
```
adracus, line 59:
```suggestion
- the target needs to know the load balancer's IP and needs to answer with it (and to receive traffic with it)
```

docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+- answers to request will be directly delivered since all details are known to the target
+- payload packages are not changed, to not lose any information
+
+## Proposal
+A network loadbalancer CRD allows the user to define the network function. all entries except the `targetNetworkInterface` must be immutable.
```
adracus, line 64:
```suggestion
A `LoadBalancer` allows the user to define the network function. `type`, `ipFamilies` and `networkRef` are immutable.
```
Regarding immutability, potentially also `ports` should be mutable.

docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+    - name: db
+      protocol: udp
+      # portrange
+      port: 1024
+      portEnd: 2048
+  #all networkInterfaces must be connected to this network.
+  networkRef:
+    name:
+  #a selector for the NetworkInterfaces to be load balancer targets, normal kubernetes selector logic
+  #TODO explicit machines ?
```
adracus, line 96: Please check / remove all TODOs.

docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+    matchLabels: 
+      key: db
+      foo: bar
+status:
+  #the ips the load balancer is listening to (max 2, an IPv4 and/or IPv6)
+  ips: 
+    - 45.86.152.88
+    - 2001::
+```
+### internal reconcile object (not customer visible)
```
adracus, line 107: We don't care about customer visibility here - that's an aspect that will be handled via RBAC.
The `LoadBalancerRouting` resource is more a resource that shows the expanded routing state at a certain point in time, hence the wording `internal` does not quite match.

docs/proposals/04-nat-gateway.md
```diff
@@ -0,0 +1,105 @@
...
+## Motivation
+Some machines of the network have no public IP addresses. But those machines also need public internet access. A NAT gateway provides this functionality by providing a default gateway to the private machines and an address translation for incoming and outgoing traffic. OnMetal needs also to provide access to private machines (e.g. downloading container images). A standard setup is to use a VirtualIp ([OEP-1](01-networking-integration.md)) but this exposes the machines with a public IP address. This service introduces NAT gateway for everyone not having a public IP address.
+
+### Goals
+- define a user facing API (onMetal api) for the NAT Gateway
+- defining the maximum outgoing connections of any machine in the network
+- defining the (public) IP Address of the gateway
+
+### Non-Goals
+- gateways are network function
```
adracus, line 43: Could you elaborate more here?

docs/proposals/04-nat-gateway.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Motivation
+Some machines of the network have no public IP addresses. But those machines also need public internet access. A NAT gateway provides this functionality by providing a default gateway to the private machines and an address translation for incoming and outgoing traffic. OnMetal needs also to provide access to private machines (e.g. downloading container images). A standard setup is to use a VirtualIp ([OEP-1](01-networking-integration.md)) but this exposes the machines with a public IP address. This service introduces NAT gateway for everyone not having a public IP address.
+
+### Goals
+- define a user facing API (onMetal api) for the NAT Gateway
+- defining the maximum outgoing connections of any machine in the network
+- defining the (public) IP Address of the gateway
```
adracus, line 40: We currently don't have the capability to define the public IP but only the capability to request 'a' public IP.

docs/proposals/04-nat-gateway.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Non-Goals
+- gateways are network function 
+
+## Proposal
+The NatGateway definition allows to define a gateway for a network. The prefix is implying the Network (VNI) this NAT-Gateway runs on. Better: every machine in that prefix will get a NAT-Gateway. If a machine gets a virtual IP it no longer needs a NAT-Gateway assignment. Anything except the `natIPs` is immutable on the NatGateway. Those are used to scale the size of the Gateway. Changes of the `PortPerMachine` cause the whole NAT-Gateway to be recomputed and can therefore only happen on initializing
```
adracus, line 46: Please remove all occurrences of `VNI` / any implementation detail.

docs/proposals/04-nat-gateway.md
```diff
@@ -0,0 +1,105 @@
...
+  # the network the NATGateway operates on
+  networkRef: 
+    name: sample-network
+  # public IP addresses the NATGateway has. Every IP has a total of 64512 (655356-1024) ports being available to machines in the NAT domain To have more ports available add more IP addresses 
+  #TODO other types of IP addresses (e.g. internal NATGateway) and pinned IP types
+  type: Public
+  natIPs:
+    - ipFamily: IPv4
+  # defines the concurrent connections per machine and target. 64 is the default (if omitted), must be a power of 2
+  PortsPerMachine: 64                    # 64, 128, 256, 512 or 1024
```
adracus, line 65:
```suggestion
  portsPerNetworkInterface: 64                    # 64, 128, 256, 512 or 1024
```

docs/proposals/04-nat-gateway.md
```diff
@@ -0,0 +1,105 @@
...
+  namespace: customer-1
+spec:
+  # the network the NATGateway operates on
+  networkRef: 
+    name: sample-network
+  # public IP addresses the NATGateway has. Every IP has a total of 64512 (655356-1024) ports being available to machines in the NAT domain To have more ports available add more IP addresses 
+  #TODO other types of IP addresses (e.g. internal NATGateway) and pinned IP types
+  type: Public
+  natIPs:
+    - ipFamily: IPv4
```
adracus, line 63:
```suggestion
  ips:
    - name: ip1
      ipFamily: IPv4
```
We'll also need names here for
* Reconciliation disambiguation
* In the future, when we support allocating a specific public IP, we have to be able to disambiguate

docs/proposals/04-nat-gateway.md
```diff
@@ -0,0 +1,105 @@
...
+  networkInterfaceSelector:
+    matchLabels:
+      key: db
+      foo: bar
+status:
+  ips:
+    - 48.86.152.12    
+  # information of Network interfaces without a virtualIP is needed
+  portsUsed: {{ PortsPerMachine * entries in NATGatewayRouting.destinations }}
+  portsAvailable: {{ ( entries in natIPs * 64512 ) - PortsUsed}}
```
adracus, line 76: This information is implicitly available via `portsUsed` and should be removed.

docs/proposals/04-nat-gateway.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+### internal reconcile object (not customer visible)
```
adracus, line 79: Same as for `LoadBalancerRouting`, customer visibility is not a concern of the API but of a cluster admin defining RBAC on top. We should rephrase the wording the same way we did for `LoadBalancerRouting`.

docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Motivation
+The onMetal network is a fully routed network. To run workloads hosted on several nodes (in Kubernetes terms equivalent to cluster IP's) and not only on a single node (node IP's) solutions like ECMP (Equal cost multi pathing) seem to be the natural solution. But ECMP has certain limitations: the amount of nodes it can be used to distribute traffic to and a hash based algorithm to pin the network flows, that changes - based on the amount of targets - its distributed pinning. 
+To avoid those limitations and to have better control over the load balancer behavior an onMetal load balancer service needs to be provided.
+
+### Goals
+- define a user facing API (onMetal api) for the load balancer
```
afritzler, line 40: Please start all lists with an upper case.

afritzler (changes requested):

docs/proposals/04-nat-gateway.md
```diff
@@ -69,7 +69,7 @@ status:
 ```
 
 ```yaml
-apiVersion: networking.onmetal.de/v1alpha1
 kind: NATGatewayRouting
```
pradumnapandit, line 85: Which object should be reconciled **NATGateway** or **NATGatewayRouting** ? 
And Does **NatGatewayRouting** gets created/updated automatically?

docs/proposals/03-loadbalancer.md
```diff
@@ -96,7 +96,7 @@ status:
 ```
 
 ```yaml
-apiVersion: networking.onmetal.de/v1alpha1
 kind: LoadBalancerRouting
```
pradumnapandit, line 113: Which object should be reconciled **LoadBalancer** or **LoadBalancerRouting**?

docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+- load balancing is used to deliver a packet addressed to the load balancer to one of its targets via the onMetal network routing
+- the load balancing will be for the target and the source transparent
+- the target needs to know the load balancers IP and needs to answer with it (and to receive traffic with it)
+- answers to request will be directly delivered since all details are known to the target
+- payload packages are not changed, to not lose any information
+
+## Proposal
+A network loadbalancer CRD allows the user to define the network function. all entries except the `targetNetworkInterface` must be immutable. 
+
+### User defined object
```
aiivashchenko, line 66: In most clouds it is possible to define and request a load balancing objects as `Service` resources with certain types like `LoadBalancer` or `NodePort` (in case of MetalLB), and with `Ingress` resource.

Probably it is worth to mention how these objects will correlate as it would allow to avoid any further compatibility and implementation issues.

docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+  networkRef:
+    name:
+  #a selector for the NetworkInterfaces to be load balancer targets, normal kubernetes selector logic
+  #TODO explicit machines ?
+  networkInterfaceSelector:
+    matchLabels: 
+      key: db
+      foo: bar
+status:
+  #the ips the load balancer is listening to (max 2, an IPv4 and/or IPv6)
```
aiivashchenko, line 102: For extensible scalability it'd be good to support multiple IPs per load balancer, i.e. supply and configure them with same set of rules.

Then, it'd be possible to put them into the RRDNS and have multiple resolvable IP addresses per endpoint.

docs/proposals/03-loadbalancer.md
aiivashchenko: I believe, yeah, considerations regarding exact implementation should be moved to the project that will handle the provisioning of load balancer service, e.g. `lblet`.

Also, it is not clear from the motivation whether ECMP will be used or not, which features would be included (e.g. will L3DSR be supported), etc.

docs/proposals/03-loadbalancer.md
```diff
@@ -0,0 +1,128 @@
...
+  ports:
+    - name: webserver  # might be optional
+      # protocols supported UDP, TCP, SCTP
+      protocol: tcp
+      # single port
+      port: 80
+    - name: db
+      protocol: udp
+      # portrange
+      port: 1024
```
aiivashchenko, line 90: I'd perefer to have for port range

```yaml
portRange: 
  begin: 10000
  end: 11000
```

docs/proposals/04-nat-gateway.md
adracus: The `onmetal-api` controllers will apply `NATGatewayRouting` from `NATGateway`.

docs/proposals/03-loadbalancer.md
adracus: The `onmetal-api` controllers will apply `LoadBalancerRouting` from `LoadBalancer`.

docs/proposals/03-loadbalancer.md
```diff
@@ -65,37 +66,47 @@ A network loadbalancer CRD allows the user to define the network function. all e
...
+  type: Public
+  ipFamilies: [ IPv4, IPv6 ]
+
+  # limits the networkinterfaces to select from
+  networkRef:
+    name: 
   ports:
-    - name: webserver
+    - name: webserver  # might be optional
       protocol: tcp
```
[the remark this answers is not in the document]
adracus: The LB is L3. Filtering will be L4 and supports multiple protocols at the same time.

docs/proposals/03-loadbalancer.md
adracus: Kubernetes `Service`s of type `NodePort` are a feature of the `Node`s running in a Cluster.
`LoadBalancer`s are inherently cloud provider specific.

docs/proposals/03-loadbalancer.md
adracus: That's exactly the non-goal we mentioned - if multiple IPs should be supported, multiple LBs have to be created.

docs/proposals/03-loadbalancer.md
adracus: We try to integrate into the k8s environment as naturally as possible and took this approach from how k8s does network policies, see
https://kubernetes.io/docs/concepts/services-networking/network-policies/#targeting-a-range-of-ports
As we already strive to be as kubernetes-like as possible, we should continue here.

docs/proposals/03-loadbalancer.md
gehoern: done

docs/proposals/03-loadbalancer.md
gehoern: tried to make it less specific but still describing the basic idea. Took your sentence @adracus

docs/proposals/03-loadbalancer.md
gehoern: done

docs/proposals/03-loadbalancer.md
gehoern: typo

docs/proposals/03-loadbalancer.md
gehoern: done

docs/proposals/03-loadbalancer.md
gehoern: yes nicer

docs/proposals/03-loadbalancer.md
gehoern: done

docs/proposals/03-loadbalancer.md
gehoern: written different, since in the last iteration everything is immutable but the networkInterfaceSelector allows for dynamic targets.

docs/proposals/03-loadbalancer.md
gehoern: done

docs/proposals/03-loadbalancer.md
gehoern: fixed

docs/proposals/03-loadbalancer.md
gehoern: hope i clarified the text, as @adracus did here

docs/proposals/03-loadbalancer.md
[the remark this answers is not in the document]
gehoern: 👍 @adracus, clarified in text

docs/proposals/03-loadbalancer.md
gehoern: I mentioned explicitly the usage of a Kubernetes ingess. hope that helpes.

docs/proposals/03-loadbalancer.md
gehoern: 👍 @adracus that is exactly the source of the definition. btw. the implicit advantage here is to have a dedicated port and a port range in the same easy definition.

docs/proposals/03-loadbalancer.md
gehoern: in theory that is a nice idea, but the idea so so nice scalable that there should be no technical reason for having multiple ip addresses (e.g. performance distribution is not an issue here) and therefore we explicitly define a single LB for a single purpose. If you want to have the same LB semantics but with a different LB IP, you can have this by defining a copy of the LB object, but you can also realise as a customer that this may not be intended

docs/proposals/04-nat-gateway.md
gehoern: rewritten

docs/proposals/04-nat-gateway.md
gehoern: removed

docs/proposals/04-nat-gateway.md
gehoern: done

docs/proposals/04-nat-gateway.md
gehoern: fixed

docs/proposals/04-nat-gateway.md
gehoern: fixed

docs/proposals/04-nat-gateway.md
gehoern: i removed it, but it does not feel intuitive. Maybe we should rethink the whole `natIP` concept and create automatically, depending on the `portsUsed` -> therefore knowing the ports needed and ips needed the amount of natIP. so we only define `type: public` and `ipFamilies` to request from. WDYT?

docs/proposals/04-nat-gateway.md
gehoern: yes fixed

docs/proposals/04-nat-gateway.md
gehoern: yes identical to LoadBalancerRouting.

adracus (changes requested):

afritzler (approved):

adracus (approved):

PR #523, opened 2022-10-27
Make `Machine` `image` field optional

afritzler:
# Proposed Changes

To support non-RAM boots we want to allow to omit the `.spec.image` field in the `Machine` spec. The image in this case will be referenced by the corresponding `Volume`.

Fixed wrong references in api-references docs.

/cc @gehoern

### DIFF
--- a/apis/compute/v1alpha1/machine_types.go
+++ b/apis/compute/v1alpha1/machine_types.go
@@ -33,8 +33,9 @@ type MachineSpec struct {
 	// MachinePoolRef defines machine pool to run the machine in.
 	// If empty, a scheduler will figure out an appropriate pool to run the machine in.
 	MachinePoolRef *corev1.LocalObjectReference `json:"machinePoolRef,omitempty"`
-	// Image is the URL providing the operating system image of the machine.
-	Image string `json:"image"`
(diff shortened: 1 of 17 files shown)


apis/compute/v1alpha1/machine_types.go
```diff
@@ -34,6 +34,7 @@ type MachineSpec struct {
 	// If empty, a scheduler will figure out an appropriate pool to run the machine in.
 	MachinePoolRef *corev1.LocalObjectReference `json:"machinePoolRef,omitempty"`
 	// Image is the URL providing the operating system image of the machine.
+	// +optional
 	Image string `json:"image"`
```
adracus, line 38: Please add `omitempty`

apis/compute/validation/machine_test.go
```diff
@@ -55,10 +55,6 @@ var _ = Describe("Machine", func() {
 			&compute.Machine{},
 			ContainElement(RequiredField("spec.machineClassRef")),
 		),
-		Entry("no image",
```
adracus, line 58: Why is this test removed? Isn't it actually verifying it's possible to have no image?

adracus (changes requested):

adracus (approved):

PR #558, opened 2022-11-25
Implement `NATGateway` controller

lukasfrank:
# Proposed Changes

Implements https://github.com/onmetal/onmetal-api/blob/main/docs/proposals/04-nat-gateway.md

### DIFF
--- a/onmetal-controller-manager/controllers/networking/natgateway_controller.go
+++ b/onmetal-controller-manager/controllers/networking/natgateway_controller.go
@@ -38,8 +38,8 @@ import (
 )
 
 const (
-	MIN_PORT int32 = 1024
-	MAX_PORT int32 = 65535
+	MinEphemeralPort int32 = 1024
+	MaxEphemeralPort int32 = 65535
 )
 
 var (
@@ -90,9 +90,9 @@ func (r *NatGatewayReconciler) assignPorts(ctx context.Context, log logr.Logger,
 	}
 
 	used := make(map[string]bool)
-	for _, dest := range natGatewayRouting.Destinations {
-		if v, found := destinations[dest.UID]; found {
-			v.IPs = dest.IPs
+	for _, destination := range natGatewayRouting.Destinations {
+		if v, found := destinations[destination.UID]; found {
+			v.IPs = destination.IPs
 			for _, ip := range v.IPs {
 				used[getMapKey(&ip)] = true
 			}
@@ -103,25 +103,25 @@ func (r *NatGatewayReconciler) assignPorts(ctx context.Context, log logr.Logger,
 		currentIp, currentPort int32 = 0, 0
 		newDestinations        []networkingv1alpha1.NATGatewayDestination
 	)
-	for _, v := range destinations {
-		log.V(2).Info("Start processing destination", "destination", v.Name)
-		if len(v.IPs) != 0 {
-			log.V(2).Info("keep previous port assignment of destination", "destination", v.Name)
-			newDestinations = append(newDestinations, *v)
+	for _, destination := range destinations {
+		log.V(2).Info("Start processing destination", "destination", destination.Name)
+		if len(destination.IPs) != 0 {
+			log.V(2).Info("keep previous port assignment of destination", "destination", destination.Name)
+			newDestinations = append(newDestinations, *destination)
 			continue
 		}
 
-		log.V(2).Info("find free port for destination", "destination", v.Name)
+		log.V(2).Info("find free port for destination", "destination", destination.Name)
 		var candidate *networkingv1alpha1.NATGatewayDestinationIP
 		for {
 			if int(currentIp) >= len(natGateway.Status.IPs) {
-				log.V(1).Info("No ports left: Skipping destination", "name", v.Name, "UID", v.UID)
-				r.Eventf(natGateway, corev1.EventTypeWarning, events.ErrorNoPortsLeft, "no ports left: skipping destination %s %s", v.Name, v.UID)
+				log.V(1).Info("No ports left: Skipping destination", "name", destination.Name, "UID", destination.UID)
+				r.Eventf(natGateway, corev1.EventTypeWarning, events.ErrorNoPortsLeft, "no ports left: skipping destination %s %s", destination.Name, destination.UID)
 				candidate = nil
 				break
 			}
 
-			port := currentPort*portsPerNetworkInterface + MIN_PORT
+			port := currentPort*portsPerNetworkInterface + MinEphemeralPort
 			candidate = &networkingv1alpha1.NATGatewayDestinationIP{
 				IP:      natGateway.Status.IPs[currentIp].IP,
 				Port:    port,
@@ -130,7 +130,7 @@ func (r *NatGatewayReconciler) assignPorts(ctx context.Context, log logr.Logger,
 
 			if _, allocated := used[getMapKey(candidate)]; !allocated {
 				used[getMapKey(candidate)] = true
-				log.V(2).Info("found port range for destination", "destination", v.Name, "start", candidate.Port, "end", candidate.EndPort, "ip", candidate.IP.String())
+				log.V(2).Info("found port range for destination", "destination", destination.Name, "start", candidate.Port, "end", candidate.EndPort, "ip", candidate.IP.String())
 				break
 			}
 
@@ -142,9 +142,9 @@ func (r *NatGatewayReconciler) assignPorts(ctx context.Context, log logr.Logger,
 		}
 
 		if candidate != nil {
-			v.IPs = []networkingv1alpha1.NATGatewayDestinationIP{*candidate}
+			destination.IPs = []networkingv1alpha1.NATGatewayDestinationIP{*candidate}
 		}
-		newDestinations = append(newDestinations, *v)
+		newDestinations = append(newDestinations, *destination)
 	}
 
 	return newDestinations, int32(len(used)), nil
@@ -174,7 +174,7 @@ func (r *NatGatewayReconciler) reconcile(ctx context.Context, log logr.Logger, n
 	}
 	log.V(1).Info("Successfully found destinations", "Destinations", destinations)
 
-	slotsPerIP := int((MAX_PORT - MIN_PORT) / portsPerNetworkInterface)
+	slotsPerIP := int((MaxEphemeralPort - MinEphemeralPort) / portsPerNetworkInterface)
 	neededIps := int(math.Ceil(float64(len(destinations)) / float64(slotsPerIP)))
 
 	if len(natGateway.Spec.IPs) < neededIps {
--- a/onmetal-controller-manager/controllers/networking/natgateway_controller.go
+++ b/onmetal-controller-manager/controllers/networking/natgateway_controller.go
@@ -106,7 +106,7 @@ func (r *NatGatewayReconciler) findNextFreeSlot(natGateway *networkingv1alpha1.N
 	}
 }
 
-func (r *NatGatewayReconciler) assignPorts(ctx context.Context, log logr.Logger, natGateway *networkingv1alpha1.NATGateway, destinations map[types.UID]*networkingv1alpha1.NATGatewayDestination, slotsPerIP int, portsPerNetworkInterface int32) ([]networkingv1alpha1.NATGatewayDestination, int32, error) {
+func (r *NatGatewayReconciler) assignPorts(ctx context.Context, log logr.Logger, natGateway *networkingv1alpha1.NATGateway, destinations map[types.UID]commonv1alpha1.LocalUIDReference, slotsPerIP int, portsPerNetworkInterface int32) ([]networkingv1alpha1.NATGatewayDestination, int32, error) {
 	natGatewayRouting := &networkingv1alpha1.NATGatewayRouting{}
 	natGatewayRoutingKey := client.ObjectKey{Namespace: natGateway.Namespace, Name: natGateway.Name}
 	if err := r.Get(ctx, natGatewayRoutingKey, natGatewayRouting); client.IgnoreNotFound(err) != nil {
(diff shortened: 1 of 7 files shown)


onmetal-controller-manager/controllers/networking/natgateway_controller_test.go
```diff
@@ -0,0 +1,278 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("NatGatewayReconciler", func() {
+	ctx := testutils.SetupContext()
+	ns := SetupTest(ctx)
+
+	It("should reconcile the natgateway and routing destinations", func() {
```
afritzler, line 36: Would be great to also have a test covering the removal of `NetworkInterface` from the network.

onmetal-controller-manager/controllers/networking/natgateway_controller_test.go
```diff
@@ -0,0 +1,278 @@
+/*
```
afritzler, line 1: Somehow the license header is different here.

onmetal-controller-manager/controllers/networking/natgateway_controller.go
```diff
@@ -0,0 +1,318 @@
...
+	"k8s.io/client-go/tools/record"
+	"k8s.io/utils/pointer"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	MIN_PORT int32 = 1024
```
afritzler, line 41: How about making those min/max fields configurable via flags and defaulting them to those values here?

onmetal-controller-manager/controllers/networking/natgateway_controller.go
```diff
@@ -0,0 +1,318 @@
...
+		return r.delete(ctx, log, natGateway)
+	}
+	return r.reconcile(ctx, log, natGateway)
+}
+
+func (r *NatGatewayReconciler) delete(ctx context.Context, log logr.Logger, natGateway *networkingv1alpha1.NATGateway) (ctrl.Result, error) {
+	return ctrl.Result{}, nil
+}
+
+func hash(ip *networkingv1alpha1.NATGatewayDestinationIP) string {
```
afritzler, line 82: Sounds more like a hashmap key you are computing here.

onmetal-controller-manager/controllers/networking/natgateway_controller.go
```diff
@@ -0,0 +1,318 @@
...
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("error finding destinations: %w", err)
+	}
+	log.V(1).Info("Successfully found destinations", "Destinations", destinations)
+
+	slotsPerIP := int((MAX_PORT - MIN_PORT) / portsPerNetworkInterface)
+	neededIps := int(math.Ceil(float64(len(destinations)) / float64(slotsPerIP)))
+
+	if len(natGateway.Spec.IPs) < neededIps {
+		log.Info(fmt.Sprintf("Got %d ips but %d are needed", len(natGateway.Spec.IPs), neededIps))
```
afritzler, line 181: I would suggest using a structured logging here instead of a nested `fmt.Sprintf`.

afritzler (changes requested):

onmetal-controller-manager/controllers/networking/natgateway_controller.go
lukasfrank: These values are more or less constant. In order to have a remainder free port distribution they need to be chosen carefully. To avoid unnecessary validation code, I would propose to postpone until it's needed.

onmetal-controller-manager/controllers/networking/natgateway_controller_test.go
```diff
@@ -0,0 +1,278 @@
...
+		ip := netip.IPv4Unspecified()
+		for _, v := range natGateway.Spec.IPs {
+			ip = ip.Next()
+			natGateway.Status.IPs = append(natGateway.Status.IPs, networkingv1alpha1.NATGatewayIPStatus{
+				Name: v.Name,
+				IP: commonv1alpha1.IP{
+					Addr: ip,
+				},
+			})
+		}
```
adracus, line 216: Rather than having logic in a test, patch constants into the `natGateway.Status` - in this case, since you know your test scenario has two IPs, just patch two constant IPs in.

onmetal-controller-manager/controllers/networking/natgateway_controller_test.go
```diff
@@ -0,0 +1,278 @@
...
+		Expect(k8sClient.Patch(ctx, natGateway, client.MergeFrom(natGatewayBase))).To(Succeed())
+
+		By("creating a network interfaces")
+
```
adracus, line 220:
```suggestion
		By("creating network interfaces to exhaust the nat gateway's IPs")

```

onmetal-controller-manager/controllers/networking/natgateway_controller.go
```diff
@@ -0,0 +1,318 @@
...
+			}
+		}
+
+		if candidate != nil {
+			v.IPs = []networkingv1alpha1.NATGatewayDestinationIP{*candidate}
+		}
+		newDestinations = append(newDestinations, *v)
+	}
+
+	return newDestinations, int32(len(used)), nil
```
adracus, line 150: We should not report any destination where `IPs` are empty - this would just be unnecessary noise.

onmetal-controller-manager/controllers/networking/natgateway_controller_test.go
```diff
@@ -0,0 +1,278 @@
...
+			Expect(k8sClient.Create(ctx, nic)).To(Succeed())
+		}
+
+		By("waiting for the nat gateway routing to be updated")
+		Eventually(func(g Gomega) {
+			Expect(k8sClient.Get(ctx, natGatewayKey, natGatewayRouting)).To(Succeed())
+			g.Expect(natGatewayRouting.Destinations).To(HaveLen(nics))
+		}).Should(Succeed())
+
+		for i := 0; i < nics; i++ {
```
adracus, line 250:
```suggestion
		By("creating more network interfaces that cannot be allocated in the nat gateway anymore")
		for i := 0; i < nics; i++ {
```

onmetal-controller-manager/controllers/networking/natgateway_controller_test.go
```diff
@@ -0,0 +1,278 @@
...
+
+			g.Expect(natGatewayRouting.NetworkRef.Name).To(BeEquivalentTo(network.Name))
+			g.Expect(natGatewayRouting.NetworkRef.UID).To(BeEquivalentTo(network.UID))
+
+			g.Expect(natGatewayRouting.Destinations).To(HaveLen(1))
+
+			g.Expect(natGatewayRouting.Destinations[0].Name).To(BeEquivalentTo(nic.Name))
+			g.Expect(natGatewayRouting.Destinations[0].UID).To(BeEquivalentTo(nic.UID))
+			g.Expect(natGatewayRouting.Destinations[0].IPs).To(HaveLen(1))
+			g.Expect(natGatewayRouting.Destinations[0].IPs[0].IP).To(BeEquivalentTo(natGateway.Status.IPs[0].IP))
```
adracus, line 129: We should not use `BeEquivalentTo` unless really necessary. `BeEquivalentTo` lets mismatching types (e.g. from type aliases) slip through undiscovered. In general `Equal` is the assertion you want to make.

onmetal-controller-manager/controllers/networking/natgateway_controller_test.go
```diff
@@ -0,0 +1,278 @@
...
+			var assigned, unassigned int
+			for _, v := range natGatewayRouting.Destinations {
+				if len(v.IPs) == 0 {
+					unassigned++
+				} else {
+					assigned++
+				}
+			}
+			g.Expect(assigned).To(BeEquivalentTo(totalSlots))
+			g.Expect(unassigned).To(BeEquivalentTo(nics*2 - totalSlots))
```
adracus, line 286: We should verify here that the first set of `NetworkInterface`s is present in the `Destination`s while the second set of `NetworkInterface`s is not present here.

onmetal-controller-manager/main.go
```diff
@@ -364,6 +365,22 @@ func main() {
...
+	if controllers.Enabled(natGatewayController) {
+		if err = onmetalapiclient.SetupNATGatewayNetworkNameFieldIndexer(context.TODO(), mgr.GetFieldIndexer()); err != nil {
+			setupLog.Error(err, "unable to setup field indexer", "field", onmetalapiclient.NATGatewayNetworkNameField)
+			os.Exit(1)
+		}
+
+		if err = (&networkingcontrollers.NatGatewayReconciler{
+			Client:        mgr.GetClient(),
+			Scheme:        mgr.GetScheme(),
+			EventRecorder: mgr.GetEventRecorderFor("nat-gateway-scheduler"),
```
adracus, line 377:
```suggestion
			EventRecorder: mgr.GetEventRecorderFor("natgateways"),
```
The name `scheduler` might be misleading since we don't schedule `NATGateway`s anywhere.
Also, that naming would be more in-line with the naming of our existing `EventRecorder`s.

adracus (changes requested):

onmetal-controller-manager/controllers/networking/natgateway_controller.go
adracus: Unresolving this since the naming is not consistent.

onmetal-controller-manager/controllers/networking/natgateway_controller.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+func (r *NatGatewayReconciler) assignPorts(ctx context.Context, log logr.Logger, natGateway *networkingv1alpha1.NATGateway, destinations map[types.UID]*networkingv1alpha1.NATGatewayDestination, slotsPerIP int, portsPerNetworkInterface int32) ([]networkingv1alpha1.NATGatewayDestination, int32, error) {
+	natGatewayRouting := &networkingv1alpha1.NATGatewayRouting{}
+	if err := r.Get(ctx, types.NamespacedName{Namespace: natGateway.Namespace, Name: natGateway.Name}, natGatewayRouting); client.IgnoreNotFound(err) != nil {
+		return nil, 0, fmt.Errorf("unable to get natgateway routing %s: %w", client.ObjectKeyFromObject(natGatewayRouting), err)
+	}
+
+	used := make(map[string]bool)
+	for _, dest := range natGatewayRouting.Destinations {
```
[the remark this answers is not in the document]
lukasfrank: This part actually reuses the `old` ip/port combination if it is still needed in order to have no change. If a target is not present anymore it will directly removed

onmetal-controller-manager/controllers/networking/natgateway_controller.go
```diff
@@ -0,0 +1,318 @@
...
+	}
+	return r.reconcile(ctx, log, natGateway)
+}
+
+func (r *NatGatewayReconciler) delete(ctx context.Context, log logr.Logger, natGateway *networkingv1alpha1.NATGateway) (ctrl.Result, error) {
+	return ctrl.Result{}, nil
+}
+
+func hash(ip *networkingv1alpha1.NATGatewayDestinationIP) string {
+	return fmt.Sprintf("%s/%d/%d", ip.IP, ip.Port, ip.EndPort)
```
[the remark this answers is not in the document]
lukasfrank: replaced through using directly `networkingv1alpha1.NATGatewayDestinationIP` as set "key"

onmetal-controller-manager/controllers/networking/natgateway_controller.go
```diff
@@ -0,0 +1,318 @@
...
+			v.IPs = dest.IPs
+			for _, ip := range v.IPs {
+				used[hash(&ip)] = true
+			}
+		}
+	}
+
+	var (
+		currentIp, currentPort int32 = 0, 0
+		newDestinations        []networkingv1alpha1.NATGatewayDestination
```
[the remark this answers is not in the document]
lukasfrank: Actually, there is only one update on the `NATGatewayRouting`. Besides that: The map returned by `findDestinations ` is used to prepare everything.

onmetal-controller-manager/controllers/networking/natgateway_controller.go
```diff
@@ -0,0 +1,318 @@
...
+	nicList := &networkingv1alpha1.NetworkInterfaceList{}
+	if err := r.List(ctx, nicList,
+		client.InNamespace(natGateway.Namespace),
+		client.MatchingLabelsSelector{Selector: sel},
+		client.MatchingFields{onmetalapiclient.NetworkInterfaceNetworkNameField: natGateway.Spec.NetworkRef.Name},
+	); err != nil {
+		return nil, fmt.Errorf("error listing network interfaces: %w", err)
+	}
+
+	destinations := map[types.UID]*networkingv1alpha1.NATGatewayDestination{}
```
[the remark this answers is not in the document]
lukasfrank: I've chosen the type `map[types.UID]*networkingv1alpha1.NATGatewayDestination` in order to save some lines of code to copy/init in the following method.

onmetal-controller-manager/controllers/networking/natgateway_controller.go
```diff
@@ -0,0 +1,318 @@
...
+				candidate = nil
+				break
+			}
+
+			port := currentPort*portsPerNetworkInterface + MIN_PORT
+			candidate = &networkingv1alpha1.NATGatewayDestinationIP{
+				IP:      natGateway.Status.IPs[currentIp].IP,
+				Port:    port,
+				EndPort: port + portsPerNetworkInterface - 1,
+			}
```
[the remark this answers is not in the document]
lukasfrank: handled better in another function

onmetal-controller-manager/controllers/networking/natgateway_controller.go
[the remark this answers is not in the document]
adracus: I wouldn't recommend doing this just to save some lines - explicit, readable code is better than short code.

onmetal-controller-manager/controllers/networking/natgateway_controller.go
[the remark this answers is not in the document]
adracus: In a single reconciliation there is only one update, however you'll have multiple reconciles and the event triggered.
This happens because during the first reconciliation, you build a (non-final) allocation list, call `Patch` which updates the `NATGatewayRouting` which then again through your `Owns` clause triggers the second reconciliation, rightening the previous 'half' reconciliation. This is not fixed.

onmetal-controller-manager/controllers/networking/natgateway_controller.go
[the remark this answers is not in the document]
lukasfrank: fixed

onmetal-controller-manager/controllers/networking/natgateway_controller.go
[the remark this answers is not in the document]
lukasfrank: As discussed: not relevant.

adracus (approved):

afritzler (approved):

PR #588, opened 2022-12-19
Object Storage Proposal

lukasfrank:


docs/proposals/05-object-storage.md
```diff
@@ -0,0 +1,147 @@
...
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Alternatives](#alternatives)
+
+## Summary
+Object storage builds the basis for many cloud applications. An Object Storage provides a simplified object 
+model for files, but has a reduced set of security and access features (non-posix). This functionality is build on top
```
adracus, line 35:
```suggestion
model for files, but has a reduced set of security and access features (non-posix). This functionality is built on top 
```

docs/proposals/05-object-storage.md
```diff
@@ -0,0 +1,147 @@
...
+model for files, but has a reduced set of security and access features (non-posix). This functionality is build on top 
+of the HTTP protocol. The current market standard is [S3](https://docs.aws.amazon.com/AmazonS3/latest/API/Type_API_Reference.html). 
+This document describes how to integrate simplified S3 buckets into the OnMetal API without taking too many details of 
+the S3 feature completeness itself.
+
+## Motivation
+Object Storage is demanded by cloud native applications, therefore, OnMetal needs to provide it for a complete solution.
+The Object Storage service should be integrated into OnMetal and use all available optimizations. It is not designed 
+to be _just_ a service on top of OnMetal. The used protocol is called [S3](https://docs.aws.amazon.com/AmazonS3/latest/API/Type_API_Reference.html)
+which introduces a storage entity called bucket. For the beginning only the bucket creation and removal is covered.
```
adracus, line 44: Please clarify what is meant with
> use all available optimizations
The pool implementor should be able to provide a performant implementation, yes, but the onmetal-api abstraction can't potentially leak through all details of the underlying implementation. I think we should redact this sentence for now.

docs/proposals/05-object-storage.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Goals
+- S3-compatible object storage implementation
+- Automatic assigned public storage endpoint (addressed via DNS)
```
adracus, line 48: We should not mention DNS here - there is an internet reachable endpoint, that should be enough for the consumer to know.

docs/proposals/05-object-storage.md
```diff
@@ -0,0 +1,147 @@
...
+from a `BucketPool` as long as it can provide the performance characteristics described in the `BucketClass`. 
+The proposed API resources are similar to `Volume`, `VolumeClass` and `Volumepool` except that a volume is a 
+block device with a specific driver.
+
+
+### Bucket
+
+A `Bucket` is a *namespaced* resource to request S3-compatible object storage. 
+`Bucket`s with the `type` `Public` set (only valid type for now) are accessible from the internet. 
+In the future, the `Bucket`s can be extended with a `networkRef` to support other types.
```
adracus, line 70: We should remove the section about potential future changes - that should be discussed & accepted in another proposal.

docs/proposals/05-object-storage.md
```diff
@@ -0,0 +1,147 @@
...
+  tps: 20Mi
+```
+[//]: # (@formatter:on)
+
+### BucketPool
+
+A `BucketPool` is a *non-namespaced* logical unit and accommodates a collection of `Bucket`s. 
+The `provider`'s id (the implementor's id) of the `BucketPool` is stated in the `providerID` field. 
+Only `Bucket`s who tolerate all the taints, will land in the `BucketPool`. `BucketClasses` which can be fulfilled by 
+the provider of the`BucketPool`, are listed in the status field `availableBucketClasses`. The `state` in the status
```
adracus, line 127:
```suggestion
the provider of the `BucketPool`, are listed in the status field `availableBucketClasses`. The `state` in the status 
```

adracus (commented):
Some minor adjustments we should bring in

docs/proposals/05-object-storage.md
```diff
@@ -0,0 +1,146 @@
...
+capabilities of the underlying storage provider are expressed via a `BucketPool`. A `Bucket` can be requested 
+from a `BucketPool` as long as it can provide the performance characteristics described in the `BucketClass`. 
+The proposed API resources are similar to `Volume`, `VolumeClass` and `Volumepool` except that a volume is a 
+block device with a specific driver.
+
+
+### Bucket
+
+A `Bucket` is a *namespaced* resource to request S3-compatible object storage. 
+`Bucket`s with the `type` `Public` set (only valid type for now) are accessible from the internet.
```
MalteJ, line 69: imho the `public` setting is an ACL setting and should probably not be configured on the bucket itself. I'd expect to set this permission via the S3 API! Please check.

docs/proposals/05-object-storage.md
```diff
@@ -0,0 +1,146 @@
...
+indicates if the pool is `Available`, `Pending` or `Unavailable`.
+
+[//]: # (@formatter:off)
+```yaml
+apiVersion: storage.api.onmetal.de/v1alpha1
+kind: BucketPool
+metadata:
+  name: ceph-object-store
+spec:
+  providerID: cephlet://pool
```
MalteJ, line 136: can we hide `ceph` in our APIs? We should use something else als providerID.

lukasfrank: The providerID is basically a reference to the implementor of the pool. In our case it's `ceph` but could be also for instance `min.io`. Do you propose to remove it completely or just use a different sample?

docs/proposals/05-object-storage.md
lukasfrank: The `type` `Public` describes if the bucket should be reachable from the internet and gives us a hint what we need to provision in the background. The `ACL` settings are done via the S3 API.

MalteJ: I don't understand what it means "reachable from the internet". Why do we need to put this into the bucket spec?

MalteJ: How is it different from the ACLs?

docs/proposals/05-object-storage.md
MalteJ: Our customers shouldn't care about what storage provider we use (here: Ceph). We should expose our storage provider as something like "onmetal-object-storage" or "onmetal-block-storage", so we can replace Ceph with a different provider if we need to without customer impact.

lukasfrank: In our lowest cluster we will find a `BucketPool` like in the sample above. In our case the `cephlet` will create this resource and sets it own provider id. Through syncing the `BucketPool` up to the customer cluster the `providerID` will be replaced through the `id` from one layer below.

lukasfrank: It's the same mechanism like for the `VolumePools`. In the root cluster we have a `VolumePool`:
```
➜  ~ k get volumepools fra3 -o yaml 
apiVersion: storage.api.onmetal.de/v1alpha1
kind: VolumePool
metadata:
  name: fra3
spec:
  providerID: partition://fra3
status:
  ...
  state: Available
  ``` 

And in the lowest cluster we can identify the actual provider: 
```
apiVersion: storage.api.onmetal.de/v1alpha1
kind: VolumePool
metadata:
  name: ceph
spec:
  providerID: cephlet://pool
status:
  availableVolumeClasses:
   ...
  state: Available
  ```

gehoern: I agree here. But currently this is an additional feature in the pool. So the providerID is just 'who cares about this pool' ... i guess for the moment those names are not to be valued too much - cephlet is an operator and exposes somewhat the relation to ceph, but storagelet was also a possible name, that would conceal this better. I would not spend to much time on provider names ... will probably change anyhow.

docs/proposals/05-object-storage.md
MalteJ: That is handled by the S3 protocol afaik. I'd like to remove this parameter if we don't know what it's for.

docs/proposals/05-object-storage.md
MalteJ: It probably doesn't matter for the spec (and is no blocker for this proposal) but we should still get this right.

docs/proposals/05-object-storage.md
lukasfrank: I checked with other implementations of S3 and agree with you @MalteJ that it's better to remove the field for now.

docs/proposals/05-object-storage.md
lukasfrank: I assume this conversation is also resolved.

adracus (approved):

PR #589, opened 2022-12-20
Fix the `VolumePoolState`

lukasfrank:
# Proposed Changes

- Rename `NotAvailable` to `Unavailable`


adracus (approved):

adracus:
@lukasfrank could you please also update the constants in the internal types?

PR #603, opened 2023-01-10
Storage encryption proposal

kasabe28:
# Proposed Changes

Storage encryption proposal to enable onmetal volume encryption with user provided encryption key


docs/proposals/06-storage-encryption.md
```diff
@@ -0,0 +1,84 @@
...
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Details](#Details)
+- [Proposal](#proposal)
+
+## Summary
+One of the important feature of Cloud Native IaS is to provide secure storage. This functionality is built on top of Ceph-CSI supported individual RBD PersistentVolumeClaim encryption, more information can be found here https://rook.io/docs/rook/v1.9/ceph-csi-drivers.html#enable-rbd-encryption-support. This proposal focuses on providing option to enable encryption for individual onmetal Volume.
```
lukasfrank, line 33: IaS -> IaaS

lukasfrank, line 33: > Ceph-CSI supported individual RBD PersistentVolumeClaim encryption, more information can be found here https://rook.io/docs/rook/v1.9/ceph-csi-drivers.html#enable-rbd-encryption-support.

this seems like an implementation detail and should not part of an API proposal

lukasfrank (changes requested):
For me is the overall flow a bit unclear. Can you elaborate the interplay of a `VolumePool` and a `Volume`. Do you implicit assume that a `VolumePool` is encrypted (that assigned volumes) if the secret in `spec.encryption` is given? What is expected if the `Volume` encryption flag is set to *false* and it is assigned to an "encrypted" pool?

docs/proposals/06-storage-encryption.md
kasabe28: done

kasabe28:
> 

As per discussion now encryption key secret reference will be provided in `Volume`.
  - If encryption enabled flag is false, then volume will not be encrypted
  - If encryption enabled flag is set to true, then user can provide encryption key via secret reference to encrypt onmetal `Volume`.
  - If encryption enabled flag is set to true and user has not provided encryption key secret reference in `Volume`, then `Volume` will be encrypted with shared key of storage provider.

kasabe28:
> For me is the overall flow a bit unclear. Can you elaborate the interplay of a `VolumePool` and a `Volume`. Do you implicit assume that a `VolumePool` is encrypted (that assigned volumes) if the secret in `spec.encryption` is given? What is expected if the `Volume` encryption flag is set to _false_ and it is assigned to an "encrypted" pool?

As per discussion now encryption key secret reference will be provided in Volume.

If encryption enabled flag is false, then volume will not be encrypted
If encryption enabled flag is set to true, then user can provide encryption key via secret reference to encrypt onmetal Volume.
If encryption enabled flag is set to true and user has not provided encryption key secret reference in Volume, then Volume will be encrypted with shared key of storage provider.

docs/proposals/06-storage-encryption.md
```diff
@@ -0,0 +1,86 @@
...
+As part of Storage encryption feature Onmetal API to support option to enable encryption of Volumes. Each `Volume` can set encryption enabled flag and provide encryption key secret reference(Optional).
+
+### Goals
+  - Allow user to enable/disable volume encryption with flag
+  - Encrypt volume with user provided encryption key
+
+### Details
+  - By default encryption enabled flag will be set to false
+  - If encryption enabled flag is set to true, then user can provide encryption key via secret reference to encrypt onmetal `Volume`.
+  - If encryption enabled flag is set to true and user has not provided encryption key secret reference in `Volume`, then `Volume` will be encrypted with shared key of storage provider.
```
ManuStoessel, line 46: should we really default to using the shared key of the storage provider or should we create a secret with a new random key for the user? Similar to the way the user would create the secret themselves we could auto-create a secret with a predictable name (e.g. same name as the `Volume`) and fill the `encryption.keySecretRef` with the name of the auto-generated secret.

ManuStoessel (commented):
Up for discussion: behaviour when a user does not set a secret for a encrypted volume explicitely. I would prefer auto-generating an individual secret instead of relying on a shared secret in that case.

docs/proposals/06-storage-encryption.md
adracus: The storage provider does not and should not offer a shared key.
We should have the encryption secret mandatory in the API (renaming to `secretRef` would be more concise & still imply the same use).
Default keys and other QoL use cases can be implemented as separate (admission / Kyverno etc.) plugins and should not be part of the API extension proposal.

kasabe28: > We should have the encryption secret mandatory in the API

Only when encryption enabled is set to true?

kasabe28: If we don't want to use any default key as of now, then we can remove encryption enabled flag and make `secretRef ` as optional

docs/proposals/06-storage-encryption.md
```diff
@@ -0,0 +1,84 @@
...
+  - Allow user to enable/disable volume encryption by providing secret reference
+  - Encrypt volume with user provided encryption key
+
+### Details
+  - User can provide encryption key via `encryption.secretRef` to encrypt onmetal `Volume`
+  - Presence of `encryption.secretRef` indicates `Volume` has to be encrypted.
+  - If `encryption.secretRef` is not provided by user, then onmetal `Volume` remains unencrypted
+
+## Proposal
+The proposal to provide storage encryption introduces new field `encryption.secretRef` in existing `Volume` type. `encryption.secretRef` is an optional field for encryption key secret reference.
```
adracus, line 49: The proposal introduces a new field `encryption` with the currently single attribute `secretRef`, referencing a secret to use for encryption.

adracus (changes requested):

docs/proposals/06-storage-encryption.md
```diff
@@ -0,0 +1,84 @@
...
+    secretRef: encryption-key-secret    # this is optional
+```
+[//]: # (@formatter:on)
+
+Secret for encryption key
+
+[//]: # (@formatter:off)
+```yaml
+apiVersion: v1
+kind: Secret
```
afritzler, line 77: Shouldn't we say a few words about the encryption secret? Like what exactly that encryption key is. Also as @adracus pointed out the fixed key might be something we want to explain.

afritzler (changes requested):

docs/proposals/06-storage-encryption.md
```diff
@@ -0,0 +1,84 @@
...
+  namespace: default
+spec:
+  volumeClassRef:
+    name: fast
+  volumePoolRef:
+    name: ceph
+  resources:
+    storage: 1Gi
+  encryption:
+    secretRef: encryption-key-secret    # this is optional
```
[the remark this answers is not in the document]
kasabe28: done by removing comment and adding explanation in proposal

docs/proposals/06-storage-encryption.md
kasabe28: done

docs/proposals/06-storage-encryption.md
```diff
@@ -0,0 +1,84 @@
...
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Details](#Details)
+- [Proposal](#proposal)
+
+## Summary
+One of the important feature of Cloud Native IaaS is to provide secure storage. This proposal focuses on providing option to enable encryption for individual onmetal Volume.
+
+## Motivation
+As part of Storage encryption feature Onmetal API to support option to enable encryption of Volumes. Each `Volume`  provide encryption key secret reference(Optional).
```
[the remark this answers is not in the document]
kasabe28: done

docs/proposals/06-storage-encryption.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Goals
+  - Allow user to enable/disable volume encryption by providing secret reference
+  - Encrypt volume with user provided encryption key
```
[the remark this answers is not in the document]
kasabe28: done.

docs/proposals/06-storage-encryption.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Details
```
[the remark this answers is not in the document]
kasabe28: done

docs/proposals/06-storage-encryption.md
```diff
@@ -0,0 +1,84 @@
...
+
+[//]: # (@formatter:off)
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: encryption-key-secret
+  namespace: default
+stringData:
+  encryptionKey: test-encryption
```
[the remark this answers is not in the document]
kasabe28: Yes, it should always `encryptionKey `as key in secret. documented it.

docs/proposals/06-storage-encryption.md
kasabe28: done

docs/proposals/06-storage-encryption.md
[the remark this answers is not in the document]
pradumnapandit: **The `secretRef` mentioned above, is DEK (Data encryption key)  or KEK(Key encryption key Passphrase)?**

**Background**:
With CephCSI the passphrase we provide is KEK and available per storage class/volumepool and CephCSI generates Unique Encryption keys(DEK) per volume
https://github.com/ceph/ceph-csi/blob/v3.6.0/docs/deploy-rbd.md#encryption-metadata-configuration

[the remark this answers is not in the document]
ushabelgur: Its Data Encryption Key not the passphrase.

adracus (approved):

afritzler (approved):

docs/proposals/06-storage-encryption.md
```diff
@@ -0,0 +1,89 @@
...
+  namespace: default
+spec:
+  volumeClassRef:
+    name: fast
+  volumePoolRef:
+    name: ceph
+  resources:
+    storage: 1Gi
+  encryption:
+    secretRef: encryption-key-secret
```
hardikdr, line 87: This should be LocalObjectRef:
```
    secretRef: 
       name: encryption-key-secret
```

adracus: Will be addressed in implementation.

kasabe28: It's already addressed in implementation.

PR #608, opened 2023-01-19
Resource Quota Proposal

adracus:
# Proposed Changes

This proposal is about introducing a `ResourceQuota` type to manage and limit resource usage in `Namespace`s.


ManuStoessel (approved):
LGTM, I like the approach with the scopeselector!

PR #642, opened 2023-02-09
Increase default NAT Gateway port range from 64 to 2048

MalteJ:
The default values of NAT gateway port ranges (currently 64) is too low for most use cases. Let's increase the number to 2048. This should be enough for the majority of use cases. Users who want to optimize this value, may reduce it to a size suitable for them.

# Proposed Changes
Increase default value for NAT gateway port range from 64 to 2048.

### DIFF
--- a/api/networking/v1alpha1/natgateway_type.go
+++ b/api/networking/v1alpha1/natgateway_type.go
@@ -22,6 +22,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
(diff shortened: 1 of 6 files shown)


adracus:
Please update the documentation here
https://github.com/onmetal/onmetal-api/blob/70b5ce08a21c9b8c2ed9551781b400a683cf6c8e/api/networking/v1alpha1/natgateway_type.go#L50
and here
https://github.com/onmetal/onmetal-api/blob/70b5ce08a21c9b8c2ed9551781b400a683cf6c8e/internal/apis/networking/natgateway_type.go#L47

PR #652, opened 2023-02-24
Add initial vnet-peering proposal

ManuStoessel:
# Proposed Changes

- add initial proposal for the VNet peering feature

### DIFF
--- a/docs/proposals/08-peering.md
+++ b/docs/proposals/08-peering.md
@@ -1,0 +1,168 @@
+---
+title: VNet Peering
+
+oep-number: 8
+
+creation-date: 2023-02-24
+
+status: implementable
+
+authors:
+
+- "@ManuStoessel"
+- "@gehoern"
+
+reviewers:
+
+- "@afritzler"
+- "@adracus"
+
+---
+
+# OEP-NNNN: Your short, descriptive title
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Alternatives](#alternatives)
+
+## Summary
+
+Network peering is a crucial feature to allow users to route traffic between two onmetal networks. This OEP introduces a new API that allows the owner of both networks to express the consent to peering the networks explicitely and extends the existing `Network` API to reflect an active peering.
+
+- peering connects whole networks together 
+(vf get both network route set announced)
+- clear separation from transit gateways
+- we need peering requests on both partners
+- peering request should have a ttl so they get removed if not matched
+
+## Motivation
+
+Without peering, users of onmetal would need to run all services via public IPs or via complicated VPN solutions if they are located in seperate networks.
+We want to implement the basic use case of allowing services running in two distinct onmetal networks to be able to reach eachother without any additional installation from the user and without exposing said services to the public internet.
+
+### Goals
+
+* Define API for requesting and acknowledging peering of two network
+* Peering request should have a TTL that allows for garbage collection of unanswered requests
+* Peering should result in one network's prefixes should be fully routeable from the other network's prefixes and vice versa (vf get both network route sets announced)
+
+### Non-Goals
+
+* We do not want to implement a transit gateway
+
+## Proposal
+
+We propose using a new API called `NetworkPeeringRequests` and extending the `Network` API with a `status` reflecting the peering. For a peering to succeed, both owners of the two `Networks` that should peer, have to create a matching `NetworkPeeringRequest`. On successful peering, the `status` of the corresponding `Network` objects should be updated with the other `Network`s details. A `NetworkPeeringRequest` will also have a set time-to-live (ttl), so the system will be able to automatically clean up vacant or unmatched `NetworkPeeringRequest`objects. Always two matching `NetworkPeeringRequests` are required to enable peering. Important: `localNetworkRef` must match to other `remoteNetworkRef` and vice versa.
+
+### The `NetworkPeeringRequest`type
+
+```yaml
+apiVersion: networking.api.onmetal.de/v1alpha1
+kind: NetworkPeeringRequest
+metadata:
+  namespace: customer-manuel
+  name: network-sample-request
+spec:
+  localNetworkRef:
+    name: manuels-network
+  remoteNetworkRef:
+    name: mysupernetwork
+    namespace: customer-andre #optional could be another network of the same namespace
+status:
+  ttl: 2023-02-15T19:00:00Z # omit if empty
+  state: -> pending because missing match
+         -> failed overlapping ip range
+         -> success when found a match and peered
+```
+
+* `localNetworkRef`: reference to the network that the creator of this `NetworkPeeringRequest` owns and intends to peer, needs to reside in the same namespace as the referenced `Network`object
+* `remoteNetworkRef`: reference to the network that should be peered with the one referenced in the `localNetworkRef`, namespace can be omited if this `Network`object resides in the same namespace as the `NetworkPeeringRequest``
+* `status.ttl`: ISO timestamp that reflects the point in time when this peering request is being deleted if the `status.state`is still `pending` or `failed` before. If `status.state`is `success`, this will be ignored and the object will not be deleted
+* `status.state`: can be either `pending`, `success` or `failed`. State is `pending` as long as there is no matching `NetworkPeeringRequest` present. State is `success`when there is a matching `NetworkPeeringRequest` present. State is `failed` when certain parameters stop the matching from succeeding, like e.g. overlapping IP ranges of the two referenced networks
+
+### The updated `Network` type
+
+```yaml
+apiVersion: networking.api.onmetal.de/v1alpha1
+kind: Network
+metadata:
+  namespace: customer-manuel
+  name: manuels-network
+status:
+  peeredNetworks:
+    - name: mysupernetwork
+      namespace: customer-andre
+```
+
+* `status.peeredNetworks`: is a list of successfully peered `Networks`
+
+### Example
+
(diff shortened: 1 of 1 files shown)


docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
+---
+title: VNet Peering
```
adracus, line 2: `VNet` is implementation detail and should be renamed.

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+- "@gehoern"
+
+reviewers:
+
+- "@afritzler"
+- "@adracus"
+
+---
+
+# OEP-NNNN: Your short, descriptive title
```
adracus, line 22: Remove placeholder

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Alternatives](#alternatives)
+
+## Summary
+
+Network peering is a crucial feature to allow users to route traffic between two onmetal networks. This OEP introduces a new API that allows the owner of both networks to express the consent to peering the networks explicitely and extends the existing `Network` API to reflect an active peering.
```
adracus, line 35: There is no 'owner' in a Kubernetes sense. If a resource representing peering exists, there should not be consent necessary.

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+
+- peering connects whole networks together 
+(vf get both network route set announced)
```
adracus, line 38: `vf` is implementation detail and should be removed.

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+- clear separation from transit gateways
```
adracus, line 39: What is the separation from transit gateway and in which factors do we differentiate? What are the characteristics we expect that are different?

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+- we need peering requests on both partners
```
adracus, line 40: What are 'partners'? Users should not be present in 'regular' Kubernetes objects (only `authorization.k8s.io` has the concept of user management).

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+- peering request should have a ttl so they get removed if not matched
```
adracus, line 41: Why have a `ttl`? In e.g. Kubernetes' `pod`s they are also only garbage collected if they are in state `Terminated` for a while which can be inferred from the `LastTransitionTime`. We should have a similar concept here.

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Motivation
+
+Without peering, users of onmetal would need to run all services via public IPs or via complicated VPN solutions if they are located in seperate networks.
```
adracus, line 45:
```suggestion
Without peering, users of onmetal would need to run all services via public IPs or via complicated VPN solutions if they are located in separate networks.
```

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+We want to implement the basic use case of allowing services running in two distinct onmetal networks to be able to reach eachother without any additional installation from the user and without exposing said services to the public internet.
+
+### Goals
+
+* Define API for requesting and acknowledging peering of two network
+* Peering request should have a TTL that allows for garbage collection of unanswered requests
+* Peering should result in one network's prefixes should be fully routeable from the other network's prefixes and vice versa (vf get both network route sets announced)
```
adracus, line 52: What is the routable set? What happens on conflicts?

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Non-Goals
+
+* We do not want to implement a transit gateway
```
adracus, line 56: How do we differentiate here?

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Proposal
+
+We propose using a new API called `NetworkPeeringRequests` and extending the `Network` API with a `status` reflecting the peering. For a peering to succeed, both owners of the two `Networks` that should peer, have to create a matching `NetworkPeeringRequest`. On successful peering, the `status` of the corresponding `Network` objects should be updated with the other `Network`s details. A `NetworkPeeringRequest` will also have a set time-to-live (ttl), so the system will be able to automatically clean up vacant or unmatched `NetworkPeeringRequest`objects. Always two matching `NetworkPeeringRequests` are required to enable peering. Important: `localNetworkRef` must match to other `remoteNetworkRef` and vice versa.
```
adracus, line 60: 1. An external object should not modify an existing different object. Either the management is complete (like e.g. in Kubernetes `Deployment` / `ReplicaSet`) or there are separate resources that reflect the join (e.g. k8s `endpoints`).
2. There should not be a `ttl`, and instead `condition`s like in other k8s objects. The `lastTransitionTime` of such a `condition` should be used for garbage collection`.
3. Why do we need an object to initiate peering with our own object? If we own two `Network`s, peering them should not require such a big overhead requiring approval between two objects.

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+kind: NetworkPeeringRequest
+metadata:
+  namespace: customer-manuel
+  name: network-sample-request
+spec:
+  localNetworkRef:
+    name: manuels-network
+  remoteNetworkRef:
+    name: mysupernetwork
+    namespace: customer-andre #optional could be another network of the same namespace
```
adracus, line 75: 1. Why does it have to be two `ref`s here? If I control the object, can't I select the networks to peer with each other using e.g. `labelSelector`s as a more comfortable vehicle?
2. The cross-namespace reference should be removed / not supported in the first version for multiple reasons:
  2.1 By having this field support cross-namespace references, we now have two responsibilities for this object (1. peering networks, 2. initiating mutual agreement between two namespace owners)
  2.2 What are the semantics on deletion? The initiated peering would break without it being reflected in the state.
  2.3 How is it ensured that the object identity (i.e. `UID`) stays the same?
  2.4 Kubernetes API conventions also does not recommend having x-namespace references: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#object-references
  2.5 How should RBAC be defined for `Role`s to allow cross namespace peering?

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+status:
+  ttl: 2023-02-15T19:00:00Z # omit if empty
+  state: -> pending because missing match
```
adracus, line 78: Please follow the api conventions: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+  peeredNetworks:
+    - name: mysupernetwork
+      namespace: customer-andre
+```
+
+* `status.peeredNetworks`: is a list of successfully peered `Networks`
+
+### Example
+
+Let's suppose Manuel wants to peer his `Network` `manuels-network` with André's `Network` `mysupernetwork`. Let's also assume that Manuel's namespace is `customer-manuel` and André's namespace is called `customer-andre`. With that we get the following two `NetworkPeeringRequest`objects reflecting a successful peering of the two `Network`objects presented at the end.
```
adracus, line 106: Please remove your own names to keep this generic. Also, remove the usage of the word 'customer' - this is about the API and namespaces, not about customers (that will eventually be there for sure, but customer / IAM can only exist if we respect k8s namespaces and boundaries correctly).

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+
+### The `NetworkPeeringRequest`type
+
+```yaml
+apiVersion: networking.api.onmetal.de/v1alpha1
+kind: NetworkPeeringRequest
+metadata:
+  namespace: customer-manuel
+  name: network-sample-request
+spec:
```
adracus, line 70: Where can I limit the exposure of one network to another? I recall that we once discussed with @MalteJ to have something like an exported `CIDR` so we can do collision checking.

docs/proposals/08-peering.md
adracus, line 60: The type should be called `NetworkPeering` as it represents the peering. The `Request` suffix does not make sense here.

docs/proposals/08-peering.md
```diff
@@ -0,0 +1,168 @@
...
+### The updated `Network` type
+
+```yaml
+apiVersion: networking.api.onmetal.de/v1alpha1
+kind: Network
+metadata:
+  namespace: customer-manuel
+  name: manuels-network
+status:
+  peeredNetworks:
```
adracus, line 97: One resource should not modify another resource. We should have a separate resource to embody the peering. See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#do-not-modify-the-referred-object

adracus (changes requested):

docs/proposals/08-peering.md
ManuStoessel: to 2: yes you're right, will think of an appropriate implementation
to 3: it might be that the user wanting to peer the two networks doesn't own both and so requires consent from the user that has the other network in their namespace (i.e. owns it)

docs/proposals/08-peering.md
ManuStoessel: see transitgw

docs/proposals/08-peering.md
ManuStoessel: ok, we'll treat the status.ttl field really as a status and not a trigger. the trigger would be the LastTransitionTime + maybe 7 days only when the state is pending or failed.

adracus:
Superseded by #676

PR #675, opened 2023-03-16
Proposal for internal load balancer

adracus:


docs/proposals/08-internal-load-balancer.md
```diff
@@ -0,0 +1,114 @@
...
+```yaml
+apiVersion: networking.api.onmetal.de/v1alpha1
+kind: LoadBalancer
+metadata:
+  name: my-loadbalancer
+spec:
+  type: Internal
+  networkRef:
+    name: my-network
+  ipFamilies: [IPv4, IPv6]
```
gehoern, line 84: Isn't this implicit by the ip address ?

docs/proposals/08-internal-load-balancer.md
```diff
@@ -0,0 +1,114 @@
...
+            name: my-lb-prefix-v6
+  networkInterfaceSelector:
+    matchLabels:
+      app: web
+  ports: # The port filtering is the same as for public load balancers
+  - protocol: TCP
+    port: 8080
+status:
+  ips:
+  - ip: "10.0.0.1"
```
gehoern, line 100: you specified a prefix in the spec, yes with the explicit limit of IPFamily.Bits. but you do reflect this here. Yes an IP has an implicit prefix length, but since we are specifying a prefix and have no IP object we problably should also specify it completely with /32 or /128.

docs/proposals/08-internal-load-balancer.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Alternatives
+
+* Target multiple `NetworkInterface`s + IPs and track (e.g. via discovery / concensus protocol)
+  which of these are available to solve the high-availability aspect. This has the drawback of
+  high implementation effort + having to choose from multiple IPs / putting the burden of choosing
+  the correct IP on a potential consumer.
+
+* Create `Machine`(s) that e.g. run `HAProxy` to target your services with. This comes with the
+  drawback of having to manage the `Machine`s / having multiple IPs for the load balancing machines
```
gehoern, line 112: also with private loadbalancers machines must be managed and know the ip address they are loadbalancing. In contrast to the virtual ip that is "not known" to the machine

docs/proposals/08-internal-load-balancer.md
```diff
@@ -0,0 +1,114 @@
...
+* Cross-Namespace consumption of the `LoadBalancer` - An internal `LoadBalancer` is only
+  available within one `Network`
+
+* Cross-Namespace IP allocation - IPs and prefixes are created and deleted in a single namespace.
+
+* Use `VirtualIP`s in a `LoadBalancer` of `type: Internal`:
+  `VirtualIP`s are always public (compare to AWS' `ElasticIP`) and their IP allocation differs
+  from allocating internal IPs (that don't have to be publicly routable / announced via ASN).
+  If a user always wants the same internal IP, this use case is already covered by specifying a
+  literal IP value or by referencing an `ipam.Prefix`.
```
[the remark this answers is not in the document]
afritzler: An internal load balancer will allocate exactly one IP address from the prefix. So you can have a prefix with the capacity of exactly one IP (/32 /128) or you could create a prefix for all of your internal load balancers e.g. "my-internal-lb-prefix".

docs/proposals/08-internal-load-balancer.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Non-Goals
+
+* Cross-Namespace consumption of the `LoadBalancer` - An internal `LoadBalancer` is only
+  available within one `Network`
+
+* Cross-Namespace IP allocation - IPs and prefixes are created and deleted in a single namespace.
+
+* Use `VirtualIP`s in a `LoadBalancer` of `type: Internal`:
+  `VirtualIP`s are always public (compare to AWS' `ElasticIP`) and their IP allocation differs
```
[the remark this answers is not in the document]
afritzler: For that problem we could envision something like `.spec.virtualIPRef` for the `Public` load balancer case. That way you could reuse a previously allocated `VirtualIP`. This is however is out of scope of this proposal.

docs/proposals/08-internal-load-balancer.md
afritzler: This is the current status information we have defined in our `LoadBalancer` type.

docs/proposals/08-internal-load-balancer.md
```diff
@@ -0,0 +1,114 @@
...
+  type: Internal
+  networkRef:
+    name: my-network
+  ipFamilies: [IPv4, IPv6]
+  ips:
+  - ip: 10.0.0.1 # It is possible to specify a literal IP
+  - ephemeral: # Or to allocate using an existing ipam.Prefix
+      prefixTemplate:
+        spec:
+          prefixRef: # The prefix length will always = IPFamily.Bits
```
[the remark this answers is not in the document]
afritzler: No check needed here as the allocated prefix length will be defaulted to `IPFamily.Bits`.

docs/proposals/08-internal-load-balancer.md
```diff
@@ -0,0 +1,114 @@
...
+we can only allocate IP addresses for `NetworkInterface`s and target them, however, as soon as the
+backing `Machine` fails, the service would become unavailable.
+
+To prevent this, we have to extend our current `LoadBalancer` type to also function internally.
+This should be done with a similar API as for the public use case but allow for the same
+flexibility with internal IPs as we have already with the `NetworkInterface` type.
+
+### Goals
+
+* Make a service running on multiple `Machine`s / `NetworkInterface`s in a single `Namespace`
```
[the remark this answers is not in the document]
adracus: IPs are not coupled with networks as written in the initial networking proposal. No verification will be done.

docs/proposals/08-internal-load-balancer.md
```diff
@@ -0,0 +1,114 @@
...
+  available behind a load-balanced IP without exposing it outside their `Network`
+
+* Manage the used internal IPs either via literals or by using `ipam.Prefix`es.
```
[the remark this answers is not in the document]
adracus: The setup described below is not using ephemeral IPs but ephemeral `ipam.Prefix`es.
It already describes explicitly how to use such a prefix.

docs/proposals/08-internal-load-balancer.md
adracus: These are IPs not prefixes. The IP can be allocated with a prefix, however, as described with the `prefixRef` and as how we currently handle it with the `NetworkInterface`, the `prefixRef` will always lead to a single ip (== IPFamily.Bits) be allocated.

docs/proposals/08-internal-load-balancer.md
adracus: This is the same structure we use for `NetworkInterface`s as described in our initial networking proposal. The fields can default each other but are both declaratively necessary.

docs/proposals/08-internal-load-balancer.md
adracus: That is not described in this scenario - the alternatives should already depict how cumbersome management without ILBs is.

adracus:
Self-approving according to the latest conversation with @gehoern and @afritzler to continue on this.

PR #676, opened 2023-03-17
Network peering proposal

adracus:


afritzler (approved):

docs/proposals/09-network-peering.md
```diff
@@ -0,0 +1,174 @@
...
+kind: Network
+metadata:
+  name: my-network-2
+  uid: 3030dcf9-f031-801b-f0f0-4fec2016e73a
+spec:
+  # Both networks have to have the semantically same peering configuration.
+  peerings:
+  - name: my-network-peering
+    networkRef:
+      name: my-network-2
```
kasabe28, line 121: I think it should be referring to `my-network-1 ` network as per above statement - "(i.e. reference each
other via `networkRef`).

adracus: Thanks for the pointer, fixed it!

PR #692, opened 2023-03-28
Encryption support for `VolumePoollet` & `VolumeBroker`

lukasfrank:
# Proposed Changes

- Encryption Secret (#603) will be brokered between the clusters


broker/volumebroker/server/volume_delete.go
```diff
@@ -33,6 +35,19 @@ func (s *Server) DeleteVolume(ctx context.Context, req *ori.DeleteVolumeRequest)
 		return nil, err
 	}
 
+	log.V(1).Info("Deleting encryption secret")
```
adracus, line 38: Please move the log into the following `if` block, otherwise it'd be logged even if nothing happened afterwards.

poollet/volumepoollet/controllers/volume_controller.go
```diff
@@ -261,6 +261,27 @@ func (r *VolumeReconciler) prepareORIVolumeClass(ctx context.Context, volume *st
...
+	if err := r.Get(ctx, encryptionSecretKey, encryptionSecret); err != nil {
+		err = fmt.Errorf("error getting volume encryption secret %s: %w", encryptionSecretKey, err)
+		if !apierrors.IsNotFound(err) {
+			return nil, false, fmt.Errorf("error getting volume encryption secret %s: %w", encryption.SecretRef.Name, err)
+		}
+		return nil, false, nil
+	}
+
+	return &ori.EncryptionSpec{
+		//Todo: null check needed?
```
adracus, line 280: What about this TODO?

poollet/volumepoollet/controllers/volume_controller.go
```diff
@@ -261,6 +261,27 @@ func (r *VolumeReconciler) prepareORIVolumeClass(ctx context.Context, volume *st
...
+	}
+
+	encryptionSecret := &corev1.Secret{}
+	encryptionSecretKey := client.ObjectKey{Name: encryption.SecretRef.Name}
+	if err := r.Get(ctx, encryptionSecretKey, encryptionSecret); err != nil {
+		err = fmt.Errorf("error getting volume encryption secret %s: %w", encryptionSecretKey, err)
+		if !apierrors.IsNotFound(err) {
+			return nil, false, fmt.Errorf("error getting volume encryption secret %s: %w", encryption.SecretRef.Name, err)
+		}
+		return nil, false, nil
```
adracus, line 276: What about an event on the volume for external users to know?

adracus (changes requested):

poollet/volumepoollet/controllers/volume_controller.go
lukasfrank: Sounds good 👍

broker/volumebroker/server/volume_delete.go
```diff
@@ -33,6 +35,19 @@ func (s *Server) DeleteVolume(ctx context.Context, req *ori.DeleteVolumeRequest)
 		return nil, err
 	}
 
+	if encryption := onmetalVolume.Volume.Spec.Encryption; encryption != nil {
```
adracus, line 38: Instead of manually deleting here, we should mark the encryption secret as controlled by the `Volume` it'll be referenced by (see e.g. how we do it with the ignition secret in [`machine_create.go`](https://github.com/onmetal/onmetal-api/blob/2205f7238d9692312dbd8bfe84460293f38b4775/broker/machinebroker/server/machine_create.go#L217)) - then you'll only need to do a `Delete` on the `Volume` and all objects will be automatically garbage collected.

broker/volumebroker/server/volume_list.go
```diff
@@ -110,6 +119,7 @@ func (s *Server) getOnmetalVolumeAccessSecretIfRequired(
 
 func (s *Server) aggregateOnmetalVolume(
 	onmetalVolume *storagev1alpha1.Volume,
+	onmetalEncryptionSecret *corev1.Secret,
```
adracus, line 122: we should follow the same flow as we do for `getSecret` + `accessSecret`.

broker/volumebroker/server/volume_list.go
```diff
@@ -132,7 +143,17 @@ func (s *Server) getAggregateOnmetalVolume(ctx context.Context, id string) (*Agg
...
+	if secret := onmetalVolume.Spec.Encryption; secret != nil {
+		if err := s.getManagedAndCreated(ctx, secret.SecretRef.Name, onmetalEncryptionSecret); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return nil, fmt.Errorf("error getting onmetal encryption secret %s: %w", secret.SecretRef.Name, err)
+			}
+			return nil, status.Errorf(codes.NotFound, "encryption secret %s not found", secret.SecretRef.Name)
+		}
+	}
+
+	return s.aggregateOnmetalVolume(onmetalVolume, onmetalEncryptionSecret, s.clientGetSecretFunc(ctx))
```
adracus, line 156: If we follow the same flow in `aggregateOnmetalVolume` as for `accessSecret`, this code block won't be necessary.

ori/testing/volume/client.go
```diff
@@ -0,0 +1,51 @@
...
+package volume
+
+import (
+	"context"
+
+	ori "github.com/onmetal/onmetal-api/ori/apis/volume/v1alpha1"
+	"google.golang.org/grpc"
+)
+
+func NewFakeRuntimeClient() *FakeRuntimeClient {
```
adracus, line 24: We should follow the same approach we have in `ori-machine`: Define a `RuntimeService` interface that is decoupled from any `grpc` internals, have a `FakeRuntimeService` that implements the `RuntimeService` interface and have a `remoteRuntime` that implements the `RuntimeService` using a GRPC client.

For the implementation in `ori-machine` see:
- [`remoteRuntime` runtime.go](https://github.com/onmetal/onmetal-api/blob/e70f23298bd5a7bcc67206886f3d6df85c73b5e8/ori/remote/machine/runtime.go#L27)
- [`RuntimeService` machine.go](https://github.com/onmetal/onmetal-api/blob/e70f23298bd5a7bcc67206886f3d6df85c73b5e8/ori/apis/machine/machine.go#L23)

ori/testing/volume/fake.go
```diff
@@ -0,0 +1,152 @@
...
+		id := hex.EncodeToString(data)
+
+		// Truncated versions of the id should not be numerical.
+		if _, err := strconv.ParseInt(id[:12], 10, 64); err != nil {
+			continue
+		}
+
+		return id[:length]
+	}
+}
```
adracus, line 51: We should use `idgen.Default` for this and not copy / paste the (in this case redundant) implementation from the fake machine runtime.

adracus (changes requested):

broker/volumebroker/server/volume.go
```diff
@@ -82,6 +83,16 @@ func (s *Server) convertOnmetalVolumeResources(resources corev1alpha1.ResourceLi
 	}, nil
 }
 
+func (s *Server) getOnmetalEncryption(volume *AggregateOnmetalVolume) *ori.EncryptionSpec {
```
adracus, line 86: This function should be called `convertOnmetalVolumeEncryption` as e.g. `convertOnmetalVolumeAccess`.

adracus (approved):
Approving already, please merge once the rename is done.

PR #732, opened 2023-04-24
Add `ResizePolicy` to `VolumeClass` type

### LINKED ISSUE #731: Allow Volume Expansion with `ResizePolicy` Field in VolumeClass
## Description

Currently, the `Volume` type in our system does not support the expansion of volumes. This issue seeks to address this limitation by introducing a new field, `ResizePolicy`, in the `VolumeClass` scheme with the values `Static` and `ExpandOnly`, to provide a flexible storage solution for users.

## Background

The current implementation of the `Volume` type is static in terms of storage size. Users have to pre-define the storage size while creating the volume, and they cannot increase the storage size later. This limitation restricts the usability of the `Volume` type and does not cater to the evolving storage needs of users.

## Use Case

A user may require more storage space as their application or project grows, which necessitates the expansion of the `Volume` type. Enabling volume expansion would allow users to modify the size of their volumes without the need for data migration, ultimately improving user experience and efficiency.

## Proposed Solution

We propose to add a new feature to the `VolumeClass` scheme that allows users to expand their volumes. The implementation could include the following steps:

1. Update the `VolumeClass` scheme to include a new field called `ResizePolicy`, with the allowed values `Static` (default) and `ExpandOnly`.
2. Modify the volume creation process to accept the `expansionPolicy` field from the associated `VolumeClass`. If not specified, the default value should be `Static`.
3. Implement a new method or API endpoint for resizing the volume. This method should validate whether the requested volume can be expanded or not (based on the `expansionPolicy` field in the `VolumeClass`) and handle the volume expansion process.
4. Update the documentation to include information about the new `ResizePolicy` field in the `VolumeClass` scheme and the volume expansion process.

## Acceptance Criteria

- The `VolumeClass` scheme should include a new field called `ResizePolicy` with the allowed values `Static` and `Expandable`.
- Users should be able to create expandable volumes by specifying the `ResizePolicy` field in the `VolumeClass` during volume creation.
- There should be a method or API endpoint to handle volume expansion for eligible volumes.
- The documentation should be updated to reflect these changes.

## Additional Notes

We should consider the potential performance implications of expanding a volume and take necessary precautions to avoid any negative impacts on the system. Additionally, it would be helpful to provide some guidelines on how to safely expand a volume without causing data loss or corruption.

Please feel free to share your thoughts, concerns, or suggestions regarding this proposed feature.

afritzler:
# Proposed Changes

Allow to control when a Volume can be resize by introducing a ResizePolicy field to the VolumeClass type. The first two supported modes are Static and ExpandOnly.

Fixes #731

### DIFF
--- a/internal/admission/plugin/volumeresize/admission_test.go
+++ b/internal/admission/plugin/volumeresize/admission_test.go
@@ -1,0 +1,163 @@
+// Copyright 2023 OnMetal authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package volumeresize_test
+
+import (
+	corev1alpha1 "github.com/onmetal/onmetal-api/api/core/v1alpha1"
+	storagev1alpha1 "github.com/onmetal/onmetal-api/api/storage/v1alpha1"
+	. "github.com/onmetal/onmetal-api/utils/testing"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	. "sigs.k8s.io/controller-runtime/pkg/envtest/komega"
+)
+
+var _ = Describe("Admission", func() {
+	ctx := SetupContext()
+	ns, volumePool := SetupTest(ctx)
+
+	var (
+		volumeClassStatic     = &storagev1alpha1.VolumeClass{}
+		volumeClassExpandOnly = &storagev1alpha1.VolumeClass{}
+	)
+
+	BeforeEach(func() {
+		By("creating an expand only VolumeClass")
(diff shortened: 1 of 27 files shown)


internal/admission/plugin/volumeresize/volumeexpansion_suite_test.go
```diff
@@ -0,0 +1,140 @@
...
+	})
+	Expect(err).NotTo(HaveOccurred())
+
+	Expect(apiSrv.Start()).To(Succeed())
+	DeferCleanup(apiSrv.Stop)
+
+	Expect(utilsenvtest.WaitUntilAPIServicesReadyWithTimeout(apiServiceTimeout, testEnvExt, k8sClient, scheme.Scheme)).To(Succeed())
+})
+
+func SetupTest(ctx context.Context) (*corev1.Namespace, *storagev1alpha1.VolumePool) {
```
adracus, line 111: `SetupTest` should not use a `ctx` but instead inside `BeforeEach` define its func as `func(ctx SpecContext)` and do a `DeferCleanup like

```go
DeferCleanup(func(ctx context.Context) error {
    return client.IgnoreNotFound(k8sClient.Delete(ctx, ns))
})
DeferCleanup(func(ctx context.Context) error {
    return k8sClient.DeleteAllOf(ctx, &storagev1alpha1.Volume{}))
})
```

internal/admission/plugin/volumeresize/admission_test.go
```diff
@@ -0,0 +1,163 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	. "sigs.k8s.io/controller-runtime/pkg/envtest/komega"
+)
+
+var _ = Describe("Admission", func() {
+	ctx := SetupContext()
```
adracus, line 32: Don't use `SetupContext` but instead use `ctx SpecContext` in your test methods.

adracus (changes requested):
Please add some validation for the resize policy value in the volume validation.

lukasfrank (approved):

PR #802, opened 2023-07-24
Add proposal for `VolumeSnapshot` resource 

### LINKED ISSUE #801: Request for Proposal: `Snapshot` Support
As of now, the project does not support snapshots, a valuable feature that facilitates tasks such as data backup, environment duplication, and disaster recovery. Snapshot support could potentially allow users to create and manage point-in-time snapshots of their workloads seamlessly.

The proposal should address how we could integrate snapshot support, ideally including the following points:

- An overview of the envisioned integration
- High-level implementation details
- Sample specifications for Kubernetes (k8s) API resources
- Any potential challenges or roadblocks that may arise during implementation

### Potential Next Steps

1. **Understand the current state**: Analyze the current `onmetal-api` structure to understand the best way to integrate snapshot support.
2. **Research**: Look into similar implementations to get ideas and understand potential pitfalls.
3. **Draft Proposal**: Begin drafting a proposal detailing how snapshot support could be integrated.
4. **Review**: Have the proposal reviewed by team members and stakeholders.
5. **Iterate**: Revise the proposal based on the feedback received.

afritzler:
# Proposed Changes

Add a proposal for creating and restoring `VolumeSnapshots`.

Fixes #801

### DIFF
--- a/docs/proposals/11-volumesnapshots.md
+++ b/docs/proposals/11-volumesnapshots.md
@@ -102,6 +102,7 @@ apiVersion: storage.ironcore.dev/v1alpha1
 kind: VolumeSnapshotContent
 metadata:
   name: example-volumesnapshotcontent
+  namespace: default
 spec:
   source:
     snapshotHandle: 1334353-234234-45435435 # Unique identifier for the snapshot in the storage provider
@@ -126,6 +127,7 @@ apiVersion: storage.ironcore.dev/v1alpha1
 kind: Volume
 metadata:
   name: restored-volume
+  namespace: default
 spec:
   volumeClassRef:
     name: example-volumeclass
--- a/docs/proposals/11-volumesnapshots.md
+++ b/docs/proposals/11-volumesnapshots.md
@@ -1,0 +1,197 @@
+---
+title: Introducing VolumeSnapshot and VolumeSnapshotContent types and Restoration Paths
+
+iep-number: 11
+
+creation-date: 2025-07-23
+
+status: implementable
+
+authors:
+
+- "@afritzler"
+
+reviewers:
+
+- "@afritzler"
+- "@lukasfrank"
+- "@balpert89"
+
+---
+
+# IEP-11: Introducing `VolumeSnapshot` and `VolumeSnapshotContent` types and Restoration Paths
+
+## Table of Contents
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [API Design](#api-design)
+  - [`VolumeSnapshot` Resource](#volumesnapshot-resource)
+  - [`VolumeSnapshotContent` Resource](#volumesnapshotcontent-resource)
+  - [`Volume` Resource Enhancement](#volume-resource-enhancement)
+- [`VolumeRuntime` enhancement](#volumeruntime-enhancement)
+- [Implementation Strategy](#implementation-strategy)
+- [Alternatives](#alternatives)
+
+## Summary
+
+We propose the introduction of two new resources to the IronCore project: `VolumeSnapshot` and `VolumeSnapshotContent`. 
+The `VolumeSnapshot` resource would allow users to take point-in-time snapshots of the content of a `Volume`. 
+The `VolumeSnapshotContent` represents the actual storage-provider-specific snapshot content. We also propose an 
+enhancement to the `Volume` resource to provide a path for restoring from a snapshot. These new resources and 
+changes would provide users with increased capabilities for data protection and disaster recovery.
+
+## Motivation
+
+The IronCore project currently provides interfaces for managing resources such as `Volumes`, `VolumeClasses` and `VolumePools`. 
+A `VolumePool` is a representation of a `VolumePool` provider that materializes the requested `Volume` in an underlying
+storage implementation. A `VolumePool` provider here announces the supported `VolumeClasses` to indicate to the user
+which performance characteristics of `Volumes` this particular `VolumePool` can offer.
+There is currently no direct way to snapshot the content of a `Volume`, nor is there a method 
+for defining different types of snapshots or restoring from a snapshot, which potentially puts users' data at risk.
+
+### Goals
+
+- Allow the user to take point-in-time snapshots of the content of a `Volume`.
+- Allow the user to restore a `Volume` from a snapshot.
+- Hide provider-specific details of snapshot management from the user.
+- For encrypted `Volumes`, reuse the existing encryption key of the `Volume` when restoring from a snapshot.
+
+### Non-Goals
+
+- Incremental snapshots of `Volumes`.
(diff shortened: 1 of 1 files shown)


balpert89:
We have discussed the PR internally and do not see a problem with going forward as-is. We do not recommend going forward with the mentioned alternatives as its either off-loading significant portions to the user (3rd party provider), complicate interaction (storage provider APIs) or restrict flexibility (adding to `Volume`).

docs/proposals/11-snapshots.md
```diff
@@ -0,0 +1,191 @@
...
+```yaml
+apiVersion: storage.api.onmetal.de/v1alpha1
+kind: Snapshot
+metadata:
+  name: example-snapshot
+spec:
+  volumeRef:
+    name: example-volume
+  snapshotClassRef:
+    name: example-snapshotclass
```
kasabe28, line 99: Can we also have encryption support for Snapshot as Volume encryption support is there?

balpert89: I would expect a transitive behavior here. If the volume is encrypted, the snapshot should be encrypted as well. Plus, the option to encrypt the snapshot explicitly.

sujeet01:
We're proposing new custom snapshot resources (`Snapshot`, `SnapshotClass`) under `ironcore` apiVersion, but other providers typically use upstream standard k8s snapshot resources (`snapshot.storage.k8s.io/v1`). Should we maintain consistency with our existing approach by using standard k8s snapshot apiVersion objects instead of creating custom ones and use them against our custom `ironcore` volume resource? 
Our CSI driver already uses standard K8s PV/PVC objects alongside our custom `ironcore` volume objects. We would then only need to update our `ironcore` Volume resource with snapshot-related fields to support this functionality

docs/proposals/11-volumesnapshots.md
```diff
@@ -0,0 +1,197 @@
...
+apiVersion: storage.ironcore.dev/v1alpha1
+kind: VolumeSnapshot
+metadata:
+  name: example-snapshot
+  namespace: default
+spec:
+  volumeRef:
+    name: example-volume
+status:
+  phase: Pending/Processing/Ready/Failed/Deleting
```
kasabe28, line 83: Can we use `state` instead of `phase` similar to other api objects?

docs/proposals/11-volumesnapshots.md
```diff
@@ -0,0 +1,197 @@
...
+
+The `VolumeSnapshotContent` resource will have the following structure:
+
+```yaml
+apiVersion: storage.ironcore.dev/v1alpha1
+kind: VolumeSnapshotContent
+metadata:
+  name: example-volumesnapshotcontent
+spec:
+  deletionPolicy: Delete
```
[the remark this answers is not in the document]
balpert89: My suggestion is to proceed with `deletionPolicy` because it really determines what happens when the parent resource is *deleted*.

Also, in later stages we may want to implement `retention` to have retained `VolumeSnapshotContent` resources cleaned up after a set amount of time, which would better fit a `retentionPolicy`.

docs/proposals/11-volumesnapshots.md
balpert89: agreed.

docs/proposals/11-volumesnapshots.md
```diff
@@ -0,0 +1,197 @@
...
+
+To support the creation and restoration of `VolumeSnapshot` resources, the `VolumeRuntime` interface will be 
+enhanced to include methods for:
+
+- Creating a `VolumeSnapshot` from a `Volume`.
+- Deleting a `VolumeSnapshot` and its associated `VolumeSnapshotContent`.
+- Restoring a `Volume` from a `VolumeSnapshot`.
+
+```protobuf
+service VolumeRuntime {
```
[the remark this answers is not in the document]
balpert89: The functions should not add too much complexity, I am voting for an implementation.

docs/proposals/11-volumesnapshots.md
```diff
@@ -0,0 +1,197 @@
...
+  // when creating a new Volume.
+  rpc CreateVolume(CreateVolumeRequest) returns (CreateVolumeResponse) {};
+
+  // CreateSnapshot will be used to create a snapshot of a Volume.
+  rpc CreateSnapshot(CreateSnapshotRequest) returns (CreateSnapshotResponse) {};
+  // DeleteSnapshot will be used to delete a snapshot and its associated content.
+  rpc DeleteSnapshot(DeleteSnapshotRequest) returns (DeleteSnapshotResponse) {};
+  // ListSnapshots will be used to list all snapshots managed by the volume provider
+  rpc ListSnapshots(ListSnapshotsRequest) returns (ListSnapshotsResponse) {};
+}
```
ushabelgur, line 164: Should we also add **listEvents** method for VolumeSnapshot ? or we can also update existing listEvent method of volumeRuntime to gather events of VolumeSnapshot

docs/proposals/11-volumesnapshots.md
afritzler: Addressed this issue.

docs/proposals/11-volumesnapshots.md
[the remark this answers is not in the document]
afritzler: I extended the runtime interface with the missing methods.

docs/proposals/11-volumesnapshots.md
afritzler: We will be re-using the `ListEvents` and then disambiguate in the implementation the correct emitting of events.

docs/proposals/11-volumesnapshots.md
```diff
@@ -0,0 +1,207 @@
...
+  volumeSnapshotRef:
+    name: example-snapshot
+    namespace: default
+    uid: 12345678-1234-5678-1234-123456789012
+```
+    
+#### Fields:
+
+- `deletionPolicy`: Defines what happens to the snapshot content when the `VolumeSnapshot` is deleted. It can be `Delete` or `Retain`.
+- `source`: Contains the `snapshotHandle`, which is a unique identifier for the snapshot in the storage provider.
```
Rohit-0505, line 118: we can remove the `deletionPolicy` from the field's description as well

kasabe28 (approved):

docs/proposals/11-volumesnapshots.md
```diff
@@ -0,0 +1,206 @@
...
+
+### `VolumeSnapshotContent` Resource
+
+The `VolumeSnapshotContent` resource will have the following structure:
+
+```yaml
+apiVersion: storage.ironcore.dev/v1alpha1
+kind: VolumeSnapshotContent
+metadata:
+  name: example-volumesnapshotcontent
```
sujeet01, line 104: `VolumeSnapshotContent` should also have `namespace` field since it's a namespace-scoped resource.

PR #811, opened 2023-07-31
Enhance Binding process

adracus:
Remove `Phase` attributes from volumes, network interfaces, networks etc. and simplify binding to be done by the 'executing' controller (e.g. `NetworkInterface` controller tries to set `targetRef` on a `VirtualIP`).

Update related code in machinepoollet and controllers.

Update test cases.


internal/controllers/compute/machine_ephemeralvolume_controller.go
```diff
@@ -0,0 +1,225 @@
...
+		volume := &storagev1alpha1.Volume{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace:   machine.Namespace,
+				Name:        volumeName,
+				Labels:      ephemeral.VolumeTemplate.Labels,
+				Annotations: ephemeral.VolumeTemplate.Annotations,
+			},
+			Spec: ephemeral.VolumeTemplate.Spec,
+		}
+		_ = ctrl.SetControllerReference(machine, volume, r.Scheme())
```
lukasfrank, line 80: is there a reason for skipping the error check?

lukasfrank: same question for the other types

adracus: The error can only occur if you're modifying an existing object (i.e. one you retrieved from the server). In this case since we're initializing the struct, we can be sure that there are no `OwnerReferences` and thus this cannot fail.

PR #816, opened 2023-08-03
VolumePool: Rename Available/Used fields to Capacity/Allocatable

lukasfrank:
# Proposed Changes

-
-
-

Fixes #


poollet/volumepoollet/controllers/controllers_suite_test.go
```diff
@@ -242,3 +260,8 @@ func SetupTest() (*corev1.Namespace, *storagev1alpha1.VolumePool, *storagev1alph
 
 	return ns, vp, vc, expandableVc, srv
 }
+
+func ExpectVolumeDeleted(ctx context.Context, volume *storagev1alpha1.Volume) {
```
afritzler, line 264: Do we want to have this method exported? I would keep it package internal.

poollet/volumepoollet/vcm/generic.go
```diff
@@ -143,28 +144,28 @@ func (g *Generic) Start(ctx context.Context) error {
 	return nil
 }
 
-func (g *Generic) GetVolumeClassFor(ctx context.Context, name string, caps *ori.VolumeClassCapabilities) (*ori.VolumeClass, error) {
+func (g *Generic) GetVolumeClassFor(ctx context.Context, name string, caps *ori.VolumeClassCapabilities) (*ori.VolumeClass, int64, error) {
```
afritzler, line 147: Do we really want to have an `int64` return value or use some `Quantity` type here?

afritzler (changes requested):

afritzler (approved):

PR #856, opened 2023-09-13
Hide resource visibility if user is not in defined group

### LINKED ISSUE #813: Limit visibility of `<Pool>.Status.Capacity / Allocatable`
# Summary

Currently, each user that can `get` / `list` a pool can see its `Status.Capacity` / `Status.Allocatable`.
Since this is potentially confidential information, we should enable hiding the field for users without the
`<pool>/stats` permission.

## Example

For a `MachinePool`, if a user doesn't have the RBAC `machinepools/stats` permission, the machine pool status should *not* include `Capacity` / `Allocatable`. If the user has this permission, the fields should be included.

lukasfrank:
# Proposed Changes

- Added `pool-status-view-allowed-groups` flag to api-server to define groups which can unhide the pool resources
- If `pool-status-view-allowed-groups` is empty every user can see pool resources

Fixes #813


internal/rbac/rbac.go
```diff
@@ -0,0 +1,43 @@
...
+	if groups == nil {
+		return true
+	}
+
+	user, ok := request.UserFrom(ctx)
+	if !ok {
+		return false
+	}
+
+	println(strings.Join(user.GetGroups(), ", "), "vs", strings.Join(groups, ", "))
```
afritzler, line 36: Do we want to really use `println` here?

internal/app/apiserver/apiserver.go
```diff
@@ -100,6 +102,9 @@ func (o *OnmetalAPIServerOptions) AddFlags(fs *pflag.FlagSet) {
 
 	fs.StringVar(&o.MachinePoolletConfig.CAFile, "machinepoollet-certificate-authority", o.MachinePoolletConfig.CAFile,
 		"Path to a cert file for the certificate authority.")
+
+	fs.StringSliceVar(&o.GroupsToShowPoolResources, "groups-pool-resources", o.GroupsToShowPoolResources,
```
afritzler, line 106: How about naming the flag `--pool-status-view-allowed-groups`?

internal/app/compute_test.go
```diff
@@ -304,4 +305,60 @@ var _ = Describe("Compute", func() {
...
+			By("checking that the resources are hidden by using GET")
+			Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(machinePool), machinePool)).To(Succeed())
+			Expect(machinePool.Status.Allocatable).To(BeNil())
+			Expect(machinePool.Status.Capacity).To(BeNil())
+
+			By("checking that the resources are hidden by using LIST")
+			machinePools := &computev1alpha1.MachinePoolList{}
+			Expect(k8sClient.List(ctx, machinePools)).To(Succeed())
+			Expect(machinePools.Items).To(HaveLen(1))
+			Expect(machinePools.Items[0].Status.Allocatable).To(BeNil())
```
afritzler, line 341: How about using `ContainElement` and `HaveFields` for those array checks.

internal/app/storage_test.go
```diff
@@ -208,6 +209,62 @@ var _ = Describe("Storage", func() {
...
+			By("checking that the resources are hidden by using GET")
+			Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(volumePool), volumePool)).To(Succeed())
+			Expect(volumePool.Status.Allocatable).To(BeNil())
+			Expect(volumePool.Status.Capacity).To(BeNil())
+
+			By("checking that the resources are hidden by using LIST")
+			volumePools := &storagev1alpha1.VolumePoolList{}
+			Expect(k8sClient.List(ctx, volumePools)).To(Succeed())
+			Expect(volumePools.Items).To(HaveLen(1))
+			Expect(volumePools.Items[0].Status.Allocatable).To(BeNil())
```
afritzler, line 244: Same as above.

afritzler (changes requested):

adracus:
Can we have this concept in a more extensible way?
* Currently, via this flag, all pools are affected - fine grained control is not possible / extending this to other resources requires lots of coding and flag-passing again
* The white-out happens after the fields are deserialized - can't we somehow enhance the individual codecs to respect custom tags in the struct definition to avoid iterating over the result objects again?

afritzler:
@adracus how about using a `SubResource` for the pool capacity information? That way we can restrict the access via k8s RBAC and don't have to fiddle around in the `Status` on a per field level who is allowed to see what.

PR #961, opened 2024-01-16
Populate `NetworkInterface` status via `machinepoollet`

afritzler:
# Proposed Changes

In the current implementation the NIC status {State,IPs} is not updated by the `machinepoollet`. This leads to the problem that a NIC which is `Available` is not being represented correctly in it's status. This PR changes the `machinepoollet` logic in a way, that it populates effective IPs to the NIC status and transitions the NIC state to `Available` if the NIC could be provisioned successfully.


poollet/machinepoollet/controllers/machine_controller.go
```diff
@@ -509,6 +510,10 @@ func (r *MachineReconciler) updateNetworkInterfaceStatus(
 				errs = append(errs, err)
 			}
 		}
+
+		if err := r.updateNetworkInterfaceStatusFromIRIStatus(ctx, nic, iriNicStatus); err != nil {
+			errs = append(errs, err)
+		}
 	}
 
 	for _, nic := range unhandledNicByUID {
```
lukasfrank, line 519: Do we here also need `r.updateNetworkInterfaceStatusFromIRIStatus` ?

afritzler: Well in that case I don't have a corresponding IRI network interface, hence can't really calculate any state/ips for status. The default `Pending` state is the way to go here.

lukasfrank (approved):

PR #1006, opened 2024-02-28
Add enhancement proposal for `Prefix` IPv6 allocation

afritzler:
# Proposed Changes

- Fix naming/identifier/short names of past proposals
- Add new proposal for IPv6 `Prefix` allocation

### DIFF
--- a/docs/proposals/11-public-prefix.md
+++ b/docs/proposals/11-public-prefix.md
@@ -1,0 +1,155 @@
+---
+title: Refactor Prefix type for IPv6 usage
+
+iep-number: 11
+
+creation-date: 2024-02-25
+
+status: implementable
+
+authors:
+
+- "@afitzler"
+
+reviewers:
+
+- "@guvenc"
+- "@MalteJ"
+
+---
+
+# IEP-11: Refactor Prefix type for IPv6 usage
+
+## Table of Contents
+
+- [IEP-11: Public Prefix](#iep-11-public-prefix)
+  - [Table of Contents](#table-of-contents)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+  - [Proposal](#proposal)
+  - [Alternatives](#alternatives)
+
+## Summary
+
+In an IPv6 environment, dynamically allocating IP prefixes is crucial for efficient address space management, ensuring 
+network security, and maintaining scalability. Allowing users to freely choose IPs and prefixes can lead to conflicts, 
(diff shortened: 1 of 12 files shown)


docs/proposals/11-public-prefix.md
```diff
@@ -0,0 +1,155 @@
...
+introduced by the [Networking Integration IEP](01-networking-integration.md), does not align with these principles.
+
+## Motivation
+
+Currently, the `NetworkInterface` allows defining `IPs` and `Prefixes` for both IP families (IPv4 and IPv6) either by 
+providing a discrete value or by using an `EphemeralPrefixSource`. The `EphemeralPrefixSource` can be derived from an 
+IPv4 or IPv6 `Prefix`, allowing for the free selection of `NetworkInterface` IPs. In the IPv4 realm, it is common 
+practice to use a local, non-publicly routable IP range, with internet connectivity ensured via a `NATGateway` for 
+egress and a `LoadBalancer` for ingress traffic. However, this approach is not applicable to IPv6.
+Here we need to ensure that either a unique local IPv6 address is being used (`fd00::/8`) or an IP or `Prefix` is derived
```
guvenc, line 50: @MalteJ Do we want to support this non-routable range ? In the same VNI, the VMs can communicate with each other.

docs/proposals/11-public-prefix.md
```diff
@@ -0,0 +1,155 @@
...
+
+```yaml
+apiVersion: networking.ironcore.dev/v1alpha1
+kind: LoadBalancer
+metadata:
+  namespace: default
+  name: my-lb
+spec:
+  type: Internal
+  ipFamilies:
```
guvenc, line 142: Loadbalancer can be also dual-stack on this API level ? I mean, the LB can be ipv4 only, ipv6 only and dual-stack.
On metalnet level each ip family will need its own metalnet object but this detail can be abstracted away on this level ?

docs/proposals/11-public-prefix.md
```diff
@@ -0,0 +1,155 @@
...
+    - IPv6
+  ips:
+    - ephemeral:
+        prefixTemplate:
+          spec:
+            parentPrefix:
+              name: my-prefix
+```
+
```
guvenc, line 151: What about NAT64 ? 
Do we need to present it on this API level ? @MalteJ 
We already have NAT Gateway object with an ipv4 address, right ? so in case NICs get assigned to that 
NAT Gateway, they will simply use the IPv4 NAT Address for their DNS64 resolved targets ?
No need to mark the NAT Gateway with some ipv6 related info on this API level ?

guvenc (commented):
Thank you Andreas. I have some comments.

PR #1026, opened 2024-03-25
Implement `Network` peering controller and extend `NetworkPeeringStatus` with `State` field

### LINKED ISSUE #1025: Extend `NetworkPeeringStatus` type with `State` field and implement network peering mechanism in `ironcore`
# Summary

- Extend the existing `NetworkPeeringStatus` type with `State` field to represent if peering state is applied or not. Newly added `State` field will be either `Initial` or `Applied`.
- Implement network peering controller to update `NetworkPeeringStatus.State` with the peering `Name` and `State` fields. Also add `PeeringClaimRefs` respectively, if network is claimed by another network. Add corresponding tests.

kasabe28:
# Proposed Changes

- Implement network peering controller
- Add tests for network peering controller
- Extend `NetworkPeeringStatus` with `State` field

Fixes #1025

### DIFF
--- a/api/networking/v1alpha1/network_type.go
+++ b/api/networking/v1alpha1/network_type.go
@@ -72,8 +72,8 @@ type NetworkState string
 type NetworkPeeringState string
 
 const (
-	// NetworkPeeringStateInitial signals that the network peering is not applied.
-	NetworkPeeringStateInitial NetworkPeeringState = "Initial"
+	// NetworkPeeringStatePending signals that the network peering is not applied.
+	NetworkPeeringStatePending NetworkPeeringState = "Pending"
 	// NetworkPeeringStateApplied signals that the network peering is applied.
 	NetworkPeeringStateApplied NetworkPeeringState = "Applied"
 )
--- a/api/networking/v1alpha1/network_type.go
+++ b/api/networking/v1alpha1/network_type.go
@@ -68,10 +68,22 @@ type NetworkStatus struct {
 // +enum
 type NetworkState string
 
+// NetworkPeeringState is the state a NetworkPeering can be in
+type NetworkPeeringState string
+
+const (
+	// NetworkPeeringStateInitial signals that the network peering is not applied.
+	NetworkPeeringStateInitial NetworkPeeringState = "Initial"
+	// NetworkPeeringStateApplied signals that the network peering is applied.
+	NetworkPeeringStateApplied NetworkPeeringState = "Applied"
+)
+
 // NetworkPeeringStatus is the status of a network peering.
 type NetworkPeeringStatus struct {
 	// Name is the name of the network peering.
 	Name string `json:"name"`
+	// State represents the network peering state
+	State NetworkPeeringState `json:"state,omitempty"`
 }
 
 const (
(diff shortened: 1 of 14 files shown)


api/networking/v1alpha1/network_type.go
```diff
@@ -68,10 +68,22 @@ type NetworkStatus struct {
 // +enum
 type NetworkState string
 
+// NetworkPeeringState is the state a NetworkPeering can be in
+type NetworkPeeringState string
+
+const (
+	// NetworkPeeringStateInitial signals that the network peering is not applied.
+	NetworkPeeringStateInitial NetworkPeeringState = "Initial"
```
afritzler, line 76: Maybe we should call this one `Pending` instead of `Initial`.

internal/apis/networking/network_type.go
```diff
@@ -68,10 +68,22 @@ type NetworkStatus struct {
 // +enum
 type NetworkState string
 
+// NetworkPeeringState is the state a NetworkPeering can be in.
+type NetworkPeeringState string
+
+const (
+	// NetworkPeeringStateInitial signals that the network peering is not applied.
+	NetworkPeeringStateInitial NetworkPeeringState = "Initial"
```
afritzler, line 76: Same as above. Please use `Pending` here.

internal/controllers/networking/network_peering_controller.go
```diff
@@ -0,0 +1,224 @@
...
+func (r *NetworkPeeringReconciler) reconcileExists(ctx context.Context, log logr.Logger, network *networkingv1alpha1.Network) (ctrl.Result, error) {
+	if !network.DeletionTimestamp.IsZero() {
+		return ctrl.Result{}, nil
+	}
+
+	return r.reconcile(ctx, log, network)
+}
+
+func (r *NetworkPeeringReconciler) reconcile(ctx context.Context, log logr.Logger, network *networkingv1alpha1.Network) (ctrl.Result, error) {
+	log.V(1).Info("Reconcile")
```
afritzler, line 47: We need to adjust the network protection controller which puts a finalizer on the network in case it has a valid paring. We could in theory do this in the peering reconciler but since we have a dedicated controller we might want to do it there.

internal/controllers/networking/network_peering_controller_test.go
```diff
@@ -0,0 +1,553 @@
...
+		network3.Status.State = networkingv1alpha1.NetworkStateAvailable
+		Expect(k8sClient.Status().Patch(ctx, network3, client.MergeFrom(baseNetwork3))).To(Succeed())
+
+		baseNetwork4 := network4.DeepCopy()
+		network4.Status.State = networkingv1alpha1.NetworkStateAvailable
+		Expect(k8sClient.Status().Patch(ctx, network4, client.MergeFrom(baseNetwork4))).To(Succeed())
+
+		By("waiting for networks to reference each other")
+		Eventually(Object(network1)).
+			Should(SatisfyAll(
```
afritzler, line 134: Please ensure that a protection finalizer is present on used networks.

afritzler (changes requested):

internal/controllers/networking/network_peering_controller.go
kasabe28: Done

afritzler (approved):

PR #1033, opened 2024-04-09
Migration of osc fixes done in onmetal-api

kasabe28:
# Proposed Changes

- Modify image build in Makefile
- Modify ChannelCapacity
- Add IntenerGateway in NetworkSpec
- Fix error aggregating ironcore nic


kasabe28:
Closed it as InternetGateway feature wont be considered for OSC ironcore migration

PR #1055, opened 2024-04-22
Add `Network` peering states `Ready`, `Error` & `Pending`

### LINKED ISSUE #1054: Change network peering states as - `Ready`, `Error` and `Pending`
# Summary

To have same states as other `ironcore` networking objects and `metalnet` objects, remove the network peering state - `Applied` and add network peering states - `Ready` and `Error` to the Network object. This will give more readability to the user.

kasabe28:
# Proposed Changes

- Add network peering states - Ready and Error
- Remove network peering state - Applied
- Add test to validate if one of the peered network is unavailable, controller continues to peer other networks.

Fixes #1054

### DIFF
--- a/internal/controllers/networking/network_peering_controller_test.go
+++ b/internal/controllers/networking/network_peering_controller_test.go
@@ -550,4 +550,101 @@ var _ = Describe("NetworkPeeringController", func() {
 		Eventually(Get(network2)).Should(Satisfy(apierrors.IsNotFound))
 		Eventually(Get(network3)).Should(Satisfy(apierrors.IsNotFound))
 	})
+
+	It("should skip network peering for that network only if network does not exist referencing a single parent network", func(ctx SpecContext) {
+		By("creating a network network-1")
+		network1 := &networkingv1alpha1.Network{
+			ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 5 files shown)


internal/controllers/networking/network_peering_controller_test.go
```diff
@@ -550,4 +550,101 @@ var _ = Describe("NetworkPeeringController", func() {
 		Eventually(Get(network2)).Should(Satisfy(apierrors.IsNotFound))
 		Eventually(Get(network3)).Should(Satisfy(apierrors.IsNotFound))
 	})
+
+	It("should skip network peering for that network only if network does not exist referencing a single parent network", func(ctx SpecContext) {
```
afritzler, line 554:
```suggestion
	It("should skip network peering for a non-existing network", func(ctx SpecContext) {
```

afritzler (changes requested):
Thanks a lot for the PR @kasabe28! The PR looks good just one minor thing.

internal/controllers/networking/network_peering_controller_test.go
kasabe28: Done

afritzler (approved):

PR #1074, opened 2024-06-17
Add validation for `NetworkPeeringClaimRef` field of `Network` type

### LINKED ISSUE #1073: Add validations for `NetworkPeeringClaimRef` fields of `Network` type
# Summary

Add validations for `NetworkPeeringClaimRef` fields of `Network` type - `Name`, `Namespace` and `UID`.

kasabe28:
# Proposed Changes

- Add validations for `PeeringClaimRefs` of `Network` type.
- Add tests for `PeeringClaimRefs` validations of `Network` type.

Fixes #1073

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -56,7 +56,6 @@ require (
 	github.com/go-openapi/jsonreference v0.20.2 // indirect
 	github.com/go-openapi/swag v0.22.3 // indirect
 	github.com/go-task/slim-sprig/v3 v3.0.0 // indirect
-	github.com/gofrs/uuid v4.4.0+incompatible
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/golang/protobuf v1.5.4 // indirect
 	github.com/google/cel-go v0.17.7 // indirect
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -56,6 +56,7 @@ require (
 	github.com/go-openapi/jsonreference v0.20.2 // indirect
 	github.com/go-openapi/swag v0.22.3 // indirect
 	github.com/go-task/slim-sprig/v3 v3.0.0 // indirect
+	github.com/gofrs/uuid v4.4.0+incompatible
```
afritzler, line 59: Can we please use this package here for UUID validation: https://github.com/google/uuid

internal/apis/networking/validation/network.go
```diff
@@ -77,6 +101,35 @@ func validateNetworkPeering(peering networking.NetworkPeering, fldPath *field.Pa
...
+	}
+
+	if peeringClaimRef.UID != "" && !isValidUID(string(peeringClaimRef.UID)) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("uid"), peeringClaimRef.UID, "invalid uid"))
+	}
+
+	return allErrs
+}
+
+func isValidUID(u string) bool {
```
afritzler, line 128: And then do something like this here:
```
func IsValidUUID(u string) bool {
    _, err := uuid.Parse(u)
    return err == nil
 }
```
We can also make this utility function public.

afritzler (changes requested):

afritzler:
Btw. we don't need to validate `types.UID` as this is a build in type and the serializer will take care of the validation.

afritzler (approved):

Issue #1070, opened 2024-05-31
Extend Runtime Interface to Support Cross-Cluster Events


lukasfrank:
# Summary

This proposal aims to extend the Runtime interface to support cross-cluster events, enhancing observability and debug-ability.

## Basic example

To support events, the following components need to be updated:
- the `IronCore Runtime Interface (IRI)`
- `<Machine/Volume/Bucket>-provider`
- `<Machine/Volume/Bucket>-poollets`

Example runtime extension: 
```GRPC
  rpc ListEvents(ListEventsRequest) returns (ListEventsResponse) {};
```

## Motivation

Currently, understanding why a resource is in a Pending state can be challenging. Dependencies such as a Volume or a NIC might introduce delays, preventing a Machine from transitioning to the Running state. Kubernetes events offer a common pattern for attaching context to objects. By introducing events and enabling synchronization between clusters, users can use `kubectl describe machine` to gain meaningful context about an object's status.

Reference:  [Kubernetes events](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/event-v1/)

ushabelgur:
This is how we are recording in events: 
for eg:
r.Eventf(machine, corev1.EventTypeNormal, events.VolumeNotReady, "Volume %s does not access information", volume.Name)

Proposed event spec: 
message EventSpec {
  string machine_id = 1; //instead of InvolvedObject(or regarding) from k8s Event can we just have machine_id? 
  string reason = 2;
  string message = 3;
  string type = 4;
  int64 event_time = 5;
  EventSeries {
      int64 count
      int64 lastObservedTime 
         }
}

These fields are deprecated in latest kubernetes version, so do we really need them ?
 first_timestamp
 last_timestamp
 count

lukasfrank:
Thanks for your proposal: As already discussed offline, let's leave the deprecated fields out our spec.

PR #1091, opened 2024-07-19
Extend Bucket Runtime Interface for Cross-Cluster Events

sujeet01:
# Proposed Changes

- Extend bucket runtime interface to support `ListEvents` method
- Implement `ListEvents` method in bucketbroker
- Add `BucketEventMapper` for periodic relisting and recording of bucket events
- Extend `irictl-bucket` with get events command
- Add tests

Part of #1070


lukasfrank (commented):
LGTM, can you maybe also add the changes from here #1093 once we merged it?

lukasfrank (approved):

PR #1092, opened 2024-07-19
Extend Volume Runtime Interface for cross-cluster Events

### LINKED ISSUE #1070: Extend Runtime Interface to Support Cross-Cluster Events

Rohit-0505:
# Proposed Changes

-  Extend volume runtime interface to support `ListEvents` method
- Add `VolumeEventMapper` for periodic relisting and recording of volume events
- Add `RBAC` rules for volume-broker events
- Add corresponding tests
- Add missing tests for `volume-broker`
- Extend `irictl-volume` with get events command

Fixes #1070

### DIFF
--- a/broker/volumebroker/server/event_list.go
+++ b/broker/volumebroker/server/event_list.go
@@ -1,0 +1,102 @@
+// SPDX-FileCopyrightText: 2024 SAP SE or an SAP affiliate company and IronCore contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package server
+
+import (
+	"context"
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/labels"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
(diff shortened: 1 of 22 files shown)


lukasfrank (commented):
LGTM, can you maybe also add the changes from here #1093 once we merged it?

lukasfrank (approved):

PR #1112, opened 2024-09-04
Upgrade K8s Dependencies to `v0.30.4`

### LINKED ISSUE #1094: Bump `k8s.io/*` deps to v0.30
# Summary

Upgrade the Kubernetes related dependencies to version v1.30. This change will also involve updating the code generation part.

## Tasks

- [x] #1113
- [ ] Bump `controller-runtime` to the latest version
- [ ] Update `hack/update-codegen.sh`
- [ ] Regenerate types

### LINKED ISSUE #1113: Bump `k8s.io/*` deps to v0.30

sujeet01:
This PR upgrades our Kubernetes package dependencies from `v0.29.4` to `v0.30.4` and implements necessary code refactoring to accommodate breaking changes. 

### Proposed Changes:

- **Upgrade K8s Dependencies**: Bumped `k8s.io/*` dependencies from `v0.29.4` to `v0.30.4` to incorporate the latest features and improvements. Fixes #1113 
- **Upgrade Controller-Runtime**: Bumped `sigs.k8s.io/controller-runtime` from `v0.17.3` to `v0.18.5` to ensure compatibility with the upgraded k8s dependencies.
- **Refactor for Breaking Changes**: Adjusted code to address breaking changes introduced in the new version.
- **Update Code Generation Script**: Adapted `hack/update-codegen.sh` script to utilize the `kube_codegen.sh` script functions, aligning with updated k8s code generation standards.
- **Update OpenAPI-Gen Package**: The `openapi-gen` has been removed from [k8s.io/code-generator](https://github.com/kubernetes/code-generator/tree/master/cmd) by [PR #123529](https://github.com/kubernetes/kubernetes/pull/123529) because it is redundant with [the version in `k8s.io/kube-openapi`](https://github.com/kubernetes/kube-openapi/tree/master/cmd/openapi-gen).
- **Update Makefile**: Updated `ENVTEST_K8S_VERSION` from `1.28.0` to `1.30.3` and performed cleanup by removing obsolete generation tool binaries.
- **Run Make Generate**: Executed `make generate` to apply all code generation tasks with the updated setup.
- **Update Import Paths**: Updated import paths to match the new locations of the generated client code.

Fixes #1094


internal/quota/evaluator/ironcore/onmetal.go
```diff
@@ -4,8 +4,8 @@
 package ironcore
```
afritzler, line 4: Can we rename this file?

.github/workflows/check-codegen.yml
```diff
@@ -14,6 +14,11 @@ jobs:
     - uses: actions/setup-go@v5
       with:
         go-version-file: 'go.mod'
+    - name: Download k8s.io/code-generator
```
afritzler, line 17: We used to install the `code-generator` via the `Makefile`:
```
.PHONY: deepcopy-gen
deepcopy-gen: $(DEEPCOPY_GEN) ## Download deepcopy-gen locally if necessary.
$(DEEPCOPY_GEN): $(LOCALBIN)
	test -s $(LOCALBIN)/deepcopy-gen || GOBIN=$(LOCALBIN) go install k8s.io/code-generator/cmd/deepcopy-gen@$(CODE_GENERATOR_VERSION)
```
I would suggest to keep it that way. If we don't need the individual generators like deepcopy etc. we can have a `code-generator` target in the `Makefile`.

afritzler (changes requested):

internal/quota/evaluator/ironcore/onmetal.go
sujeet01: Yes, we can. The package name has already been updated, so renaming the file won’t cause any issues. I’ve also renamed the other two files that were using the older name.

.github/workflows/check-codegen.yml
sujeet01: I think this additional step is no longer needed, so I've removed it from the workflow.

afritzler:
For me `make generate` fails on an clean project with:
```
./hack/update-codegen.sh: line 12: /kube_codegen.sh: No such file or directory
```

I would suggest to modify the `hack` script to fetch it from the correct location in the `go mod` cache.

sujeet01:
> For me `make generate` fails on an clean project with:
> 
> ```
> ./hack/update-codegen.sh: line 12: /kube_codegen.sh: No such file or directory
> ```
> 
> I would suggest to modify the `hack` script to fetch it from the correct location in the `go mod` cache.

Thanks for catching this issue. I have addressed this in the latest commit by reordering the steps in the `hack` script to download dependencies before setting `CODE_GEN_DIR`. This should resolve the problem, and `make generate` should now work as expected on a clean project.

afritzler (approved):

PR #1120, opened 2024-09-09
Fix wrong `restconfig` reference when using a `bootstrap` kubeconfig

nitrocb:
# Proposed Changes

- Changes to the correct restconfig in case a bootstrap-kubeconfig is used
- Prevents panic in case only bootstrap-kubeconfig is used to generate real kubeconfig


afritzler (approved):
@nitrocb thanks a lot!

PR #1128, opened 2024-09-30
Reservation-Based Scheduling Proposal for `Machine`s

### LINKED ISSUE #1277: Design Proposal for Reservation-Based Scheduling
## Objective

Introduce a Reservation-based scheduling system that enables robust, hierarchical, and dynamic resource allocation across multiple layers in the Ironcore infrastructure. The goal is to decouple resource discovery and commitment by using an intermediate `Reservation` object to enhance flexibility, reduce complexity, and support richer scheduling decisions.

## Proposed API Changes

- Extend the `IRI`
- Introduce a new `Reservation`

lukasfrank:
# Proposed Changes

- This proposal discusses possible solutions to improve the IronCore scheduling

fixes: #1277


balpert89:
+1 for the proposal. This addresses a lot of current pain points in the stack:

- better utilization for dedicated machine pools
- a transparent decision making process how a `Machine` resource ended up on a given `MachinePool`
- adding custom resources for scheduling decisions such as `EPC Memory`

Going into more detail. The current proposed approach follows a decentralized solution whereas the centralized one (in the alternatives section) follows the network stack solution. There are some disadvantages with a clustered solution, such as lack of guaranteed availability. For networking this is okay because if a critical infrastructure component is down, networking is affected anyway. The scheduling part should not have such a big impact for computing - using the decentralized approach would isolate impact on a pool level. 
Another aspect with a centralized solution is you have to deal with a lot of "boilerplate" challenges such as "eventual consistency", giving room for possible race conditions. The `Reservations` solution solves this pretty elegantly because you can introduce a time peroid until a scheduler waits for its decision and only takes the pools into consideration it finds in the status slice. Therefore, my vote goes with a decentralized approach here.

On the topic of the "scheduling decision". The reservation system is meant to have a decision *who* can provide the requested resources. Another controller can then use this to actually decide which one of those pools to actually use for the the `Machine` resource. This enables a similar behavior compared with `Node` <-> `Pod` scheduling in vanilla Kubernetes. Another point to consider is that you can enable "system" reservations to accomodate for resources that are exclusively reserved for system applications.

Some aspects that are unclear for me:
- who decides the `rating` on a given status entry, is this similar to a `priority`? How does this influence the scheduling decision?
- how will arbitrary resources be announced, such as the already mentioned `EPC Memory`? Or e.g. dedicated graphics cards?

lukasfrank:
> * who decides the `rating` on a given status entry, is this similar to a `priority`? How does this influence the scheduling decision?

Only the `pool provider` can calculates the rating (since it's the component to check if the reservation can be fulfilled)  and it is a metric on "how good the `Reservation` fits onto the related pool". It should be understood as a hint for the scheduler to take the decision. 

> * how will arbitrary resources be announced, such as the already mentioned `EPC Memory`? Or e.g. dedicated graphics cards?

In the distributed approach: There is no need anymore for announcing resources. The resource "owner" (the `pool provider`) is in charge of taking or rejecting the reservation and needs to keep track of all the resources. If arbitrary resources aren't available on a specific host, the reservation will be declined.

balpert89:
The `rating` part is clear for me now, thanks for addressing.

> In the distributed approach: There is no need anymore for announcing resources. The resource "owner" (the pool provider) is in charge of taking or rejecting the reservation and needs to keep track of all the resources. If arbitrary resources aren't available on a specific host, the reservation will be declined.

Does that mean we will deprecate the `allocatable` / `available` (https://github.com/ironcore-dev/ironcore/blob/main/api/compute/v1alpha1/machinepool_types.go#L30-L33) fields as they are not required anymore?

lukasfrank:
> The `rating` part is clear for me now, thanks for addressing.
> 
> > In the distributed approach: There is no need anymore for announcing resources. The resource "owner" (the pool provider) is in charge of taking or rejecting the reservation and needs to keep track of all the resources. If arbitrary resources aren't available on a specific host, the reservation will be declined.
> 
> Does that mean we will deprecate the `allocatable` / `available` (https://github.com/ironcore-dev/ironcore/blob/main/api/compute/v1alpha1/machinepool_types.go#L30-L33) fields as they are not required anymore?

Correct, there would be no need for this fields anymore. In case if it's used to aggregate the resources of the entire infrastructure, we can offer metrics and aggregate it the kubernetes way.

afritzler:
Can we maybe rename the proposal and the PR. Multi-level scheduling seems to be to generic.

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,133 @@
+---
+title: Reservation-Based Scheduling
+
+oep-number: 11
```
afritzler, line 4: Please change oer to iep here and in all existing proposals.

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,133 @@
...
+
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+- [Alternatives](#alternatives)
+
+## Summary
+Scheduling resources is a central process to increase utilization of cloud infrastructure. It not only involves the process to find any feasible `pool` but rather, if possible, to find an optimal compute `pool`.
```
afritzler, line 34: Instead of refering to a general `pool` I would pin this down to `MachinePools` only.

afritzler: Also, what does "optimal" mean in this context? Should we rather say here that we find a `MachinePool` which can fullfil the resource request of a `Machine` and is able to materialize a `Machine`.

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,133 @@
...
+
+
+## Motivation
+
+With `ironcore` a multi-level hierarchy can be built to design availability/failure zones. `Machine`s can be created on every layer, which are subsequently scheduled on the compute `Node`s  in the hierarchy below. A scheduler on every layer should assign a `Machine` on a `Pool` which can fulfill it. This proposal addresses the known limitations of the current implementation:
```
afritzler, line 39: Instead of compute `Node` I would refer here to compute `Pools` or `MachinePools`.

afritzler: Also what are the current limitations of the implementation. We should list them here and point out that this proposal wants to address those.

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,133 @@
...
+
+- Dynamic (`NetworkInterfaces`, `LocalDiskStorage`) and static resources (resources defined by a `Class` like `CPU`, `Memory`) should be taken into the scheduling decision.
+- Scheduling should be robust against data race conditions
+- `Machine`s on every level should be scheduled on a `Pool`
+- It should be possible to add resources influencing the scheduling decision without API change
+- Resource updates of `Pool`s should be possible
+- Adding/removing of compute `Pool`s should be possible
+
+### Non-Goals
+- The scheduler should not act on `Machine` updates like `NetworkInterface` attachments
```
afritzler, line 53: This proposal does not cover the scheduling of `Volumes`.

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Proposal
+
+For every created `Machine` a resource `Reservation` object will be created. The `poollet` will continue to broker only `Machine`s with a `.spec.machinePoolRef` set. The `Reservation` is a condensed derivation of a `Machine` containing the requested resources like `NetworkInterfaces` and the attached `MachineClass` resource list. The `IRI` will be extended to be able to broker the `Reservation`s. Once the `Reservation` hits a pool provider, the decision can be made if the `Reservation` can be fulfilled or not. In case of accepting the `Reservation`, resources needs to be blocked until a) the reservation is being deleted b) the corresponding `Machine` is being placed on the pool provider. The status of the `Reservation` is being propagated up to the layer where it was created. After some time it contains a list of possible pools where the scheduler can pick one, set the `Machine.spec.machinePoolRef` and the `poollet` will broker the `Machine` to the next level.
```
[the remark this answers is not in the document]
afritzler: Instead of "after some time" I would rather state that as soon as a root (origin) `Reservation` as an available `MachinePool` in it's status, this pool will be picked by the scheduler for assigning the `.spec.machinePoolRef` of a `Machine`.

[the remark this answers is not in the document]
afritzler: Instead of focusing on the brokering aspect I would rather write that the `machinepoollet` will call the `CreateMachine` IRI interface method to create the `Machine` against the IRI interface implementor. This can result either in a brokering of a `Machine` object from one cluster to another or the creation of a VM in case the IRI implementor is e.g. the `libvirt-provider` or any other concrete compute provider implementation.

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,133 @@
...
+    cpu: 2
+status:
+  # Pools describe where the resource reservation was possible
+  pools: 
+    - name: poolA
+      rating: 2
+      state: accepting
+    - name: poolB
+      ratring: 1
+      state: rejecting
```
afritzler, line 85: Can we come up here with better states than: accepting/rejecting?

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,133 @@
...
+
+```protobuf
+rpc ListReservations(ListReservationsRequest) returns (ListReservationsResponse) {};
+rpc CreateReservation(CreateReservationRequest) returns (CreateReservationResponse) {};
+rpc DeleteReservation(DeleteReservationRequest) returns (DeleteReservationResponse) {};
+```
+
+
+### Detailed flow
+1. If `.spec.machinePoolRef` is not set, the scheduler creates a `Reservation` that includes the required resources.
```
afritzler, line 98: Creates a `Reservation` for all available pools in the particular cluster.

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,133 @@
...
+2. `poollet`s which match `.spec.pools` of `Reservation` pick it up and broker it one layer down
+3. `poriver` evaluates the `Reservation` and sets its state to `accepting` or `rejecting`, which is then propagated up the hierarchy.
```
afritzler, line 100: Typo `provider`.

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,133 @@
...
+4. The scheduler will pick a `pool` on every layer and update the `.spec.machinePoolRef` of the related `Machine`
+5. Once a `Machine` reaches a `provider` with a relating `Reservation`, the `Reservation` will be replaced through 
+   the `Machine`. Deletion process of `Reservation` is initiated through setting state to `releasing`, which is 
+   propagated up.
+6. Scheduler will delete top `Reservation` which triggers deletion of all brokered `Reservation`s
```
afritzler, line 105: When does it delete the root (origin) `Reservation`?

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,133 @@
...
+4. The scheduler will pick a `pool` on every layer and update the `.spec.machinePoolRef` of the related `Machine`
+5. Once a `Machine` reaches a `provider` with a relating `Reservation`, the `Reservation` will be replaced through 
+   the `Machine`. Deletion process of `Reservation` is initiated through setting state to `releasing`, which is 
+   propagated up.
+6. Scheduler will delete top `Reservation` which triggers deletion of all brokered `Reservation`s 
+
+
+
+### Advantages
+- `Reservation`s can also be used to block resources for a specific use-case
```
afritzler, line 110: Without creating a `Machine` object?

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,133 @@
...
+- In case of a partial outage/network issues `Machine`s can be placed on other `Pool`s
+- `Pool`s do not leak resource information, owner of resources decides if `Reservation` can be fulfilled (less complexity for over provisioning)
+
+### Disadvantages
+- One more resource (`Reservations`) is being introduced
```
afritzler, line 115: Increases complexity in the scheduling by introducing a new resource (`Reservation`) and the scheduler flow has to be extended.

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,133 @@
...
+- Scheduling in central place is simpler
+- Availability/failure zones can be modeled in central cluster
+- (Scheduling decision takes potentially less time?)
+
+### Disadvantages
+- Resources are populated to central place and consistency needs to be guaranteed
+  - Updates are harder (failure of provider, overbooking)
+- Bookkeeping of resources needs to happen twice: in provider and central place 
+- Layered structure of hierarchy needs to be duplicated at central place
+- If central cluster is not reachable, no `Machine` can be placed
```
afritzler, line 133: No easy way to dynamicallt change pool hirarchy.

afritzler (changes requested):

docs/proposals/11-scheduling.md
balpert89: i would imagine this can be used for system components other than `Machine`s. Though I am not sure if other sub-systems would be a better fit to leverage this.

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,150 @@
...
+process to find any feasible `MachinePool` but rather, if possible, to find a `MachinePool` where the `Machine` can 
+be materialized.
+
+
+## Motivation
+
+With `ironcore` a multi-level hierarchy can be built to design availability/failure zones. `Machine`s can be created 
+on every layer, which are subsequently scheduled on `MachinePool`s  in the hierarchy below.  A scheduler on every 
+layer should assign a `Machine` on a `MachinePool` which can fulfill it.  This proposal addresses the known 
+limitations of the current implementation:
```
afritzler, line 44: No. 1 is a limitation No. 2 is an implementation detail which we have. The biggest limitation however is that we can not guarantee at the moment that we find enough resources on a compute pool and blindly schedule `Machine` objects onto pools even though they might not be fullfilled.

afritzler (changes requested):

docs/proposals/11-scheduling.md
lukasfrank: > No. 2 is an implementation detail which we have 
which is a limitation since we at the moment only support CPU and memory

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,162 @@
...
+## Proposal
+
+For every created `Machine` with an empty `spec.machinePoolRef`, the scheduler will create a resource `Reservation`. 
+The `poollet` will continue to broker only `Machine`s with a `.spec.machinePoolRef` set. The `Reservation` is a 
+condensed derivation of a `Machine` containing the requested resources like `NetworkInterfaces` and the attached 
+`MachineClass` resource list. The `IRI` will be extended to be able to manage the `Reservation`s. Once the 
+`Reservation` hits a pool provider (e.g. `libvirt-provider`), the decision can be made if the `Reservation` can be 
+fulfilled or not. In case of accepting the `Reservation`, resources needs to be blocked until: 
+1. the reservation is being deleted 
+2. the corresponding `Machine` is being placed on the pool provider.
```
balpert89, line 72: the order of those points should be switched because the `machine` is essentially replacing the `reservation`.

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,162 @@
...
+
+The status of the `Reservation` is being propagated up to the layer where it was created. As soon as the root
```
balpert89, line 74: how is it ensured that a scheduler on a lower hierarchy level does not start scheduling / acting upon the update for the particular `reservation` object? the status is consecutively brokered up so an intermediate layer will see an update to the `reservation` before the "root" `reservation` / scheduler.

balpert89: ah forget this, as soon as i hit "add" i realized there is no `machine` object on that intermediate layer because brokering for the `machine` only starts when it has a machinepool reference set.

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+
+### Detailed flow
+
+The flow to assign a `Machine` to a `Pool` consists of 3 Phases:  
+
+#### 1. Reservation Flow
+1. If `.spec.machinePoolRef` is not set, the scheduler creates a `Reservation` that includes the required resources.
+2. `poollet`s which match `.spec.pools` of `Reservation` pick it up and broker it one layer down
```
balpert89, line 120: when `.spec.pools` is left empty all pools are considered?

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,162 @@
...
+   will pick one of the `Accepted` pools of the `Reservation.status.pools` and updates the `Machine.spec.
+   machinePoolRef`.
+2. `poollet` picks up the `Machine` since `.spec.machinePoolRef` is set
+3. Once a `Machine` reaches a `provider` with a relating `Reservation`, the `Reservation` will be replaced through
+   the `Machine`. `Reservation` status will be updated to `Bound` and `poollet`s pulls the status up until to the 
+   root `Reservation`.
+
+#### 3.  Cleanup Flow
+1. The scheduler deletes root `Reservation` if `Machine` has `.spec.machinePoolRef` and if `Reservation` has a 
+   `Bound` state.
```
balpert89, line 135: should we describe the possible states for a `reservation` and also for the decision of a provider in `.status.pools[].state`?

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,162 @@
...
+- Availability/failure zones can be modeled in central cluster
+- (Scheduling decision takes potentially less time?)
+
+### Disadvantages
+- Resources are populated to central place and consistency needs to be guaranteed
+  - Updates are harder (failure of provider, overbooking)
+- Bookkeeping of resources needs to happen twice: in provider and central place 
+- Layered structure of hierarchy needs to be duplicated at central place
+- If central cluster is not reachable, no `Machine` can be placed
+- No easy way to dynamicallt change pool hirarchy.
```
balpert89, line 162: typos

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,162 @@
...
+Another solution: The providers (in the lowest layer) announce there resources to a central cluster. In these clusters a `shallow` pool represent the actual compute pool. In fact the problem of scheduling across multi-levels is transformed in a one-layer scheduling decision.  Pools between the root cluster and the leaf clusters are announced but only to represent the hierarchy and not for the actual scheduling.
+
+If a `Machine` is created, a controller creates a related `Machine` in the central cluster where the `Pool` is being picked. The `Pool` is synced back and the`.spec.machinePoolRef` being set. The `poollet` syncs the `Machine` one layer down. In every subsequent layer the path is being looked up in the central cluster until the provider is hit.
+
+### Advantages
+- Scheduling in central place is simpler
+- Availability/failure zones can be modeled in central cluster
+- (Scheduling decision takes potentially less time?)
+
+### Disadvantages
```
balpert89, line 156: another point here is that to model this correctly a poollet has to be aware of if its a leave or a branch poollet. This is required here to correctly model the topology as a branch poollet needs to announce the grouped machinepools under its own aggregated pool.

lukasfrank: I think it's covered with this point, right? `- Layered structure of hierarchy needs to be duplicated at central place`

balpert89: the importance is that it requires a change in the poollet / broker design that could potentially break the current approach for those components.

docs/proposals/11-scheduling.md
```diff
@@ -0,0 +1,133 @@
...
+  # Resources defines the requested resources in a ResourceList
+  resources:
+    localStorage: 10G
+    nics: 2
+    cpu: 2
+status:
+  # Pools describe where the resource reservation was possible
+  pools: 
+    - name: poolA
+      rating: 2
```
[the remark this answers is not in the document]
lukasfrank: As discussed offline: We will keep this out for now since the aggregation of the `rating` is not defined until now. Furthermore we might can solve this issue with a SGX-`MachineClass`. But let's revisit the topic of "optimal" scheduling once we gained some experience with the new approach.

afritzler (commented):
The propsal looks good from my side. As we are only covering the `MachineRuntimeInterface` maybe we should reflect that in the title of the proposal.

afritzler (approved):

PR #1133, opened 2024-10-10
Add e2e example for `Machine` creation using non ephemeral resources

### LINKED ISSUE #1129: E2e examples
# Summary

Create a library of sample e2e tests in the `config/samples/e2e` directory covering the following scenarios. Make sure those samples can be customized via `kustomization` e.g. for using custom `MachineClasses` and `VolumeClasses`.

## Scenarios

- [x] Create simple `Machine` with **non-ephemeral** `Volume` and `NetworkInterface` to install nginx and expose it to the internet via a `VirtualIP`
- [x] Create simple `Machine` with **ephemeral** `Volume` and `NetworkInterface` to install nginx and expose it to the internet via a `VirtualIP`
- [x] IPAM examples illustrating the `Prefix` allocation for both IPv4 and IPv6. The e2e samples should also how the sub `Prefix` allocation.
- [x] Create example illustrating `LoadBalancer` with 2 simple `Machines`, `NetworkInterface` pointing to same Network and `LoadBalancer` with label selector for created `NetworkInterface`
- [x] Create simple example illustrating 'Network' peering with two `Network`s peered to each other
- [ ] Illustrate `NetworkPolicy` example to allow ssh at port 22 to a machine from another machine

Rohit-0505:
# Proposed Changes

- Add manifests for `machine` creation and expose it to internet via `virtualIP`
- Add `ignition.yaml` to create the ignition secret

Fixes #1129(partially)


afritzler (changes requested):
Can we move the examples into a `bases` folder under `samples/e2e` and leave the patch samples out of his all together as they have to be defined by the consumer.

Rohit-0505:
> Can we move the examples into a `bases` folder under `samples/e2e` and leave the patch samples out of his all together as they have to be defined by the consumer.

done

config/samples/e2e/machine-with-non-ephemeral-resource/README.md
```diff
@@ -0,0 +1,63 @@
+# Machine Sample deployement
```
afritzler, line 1:
```suggestion
# `Machine` with non ephemeral resources 
```

config/samples/e2e/machine-with-non-ephemeral-resource/README.md
```diff
@@ -0,0 +1,63 @@
...
+- IronCore `Volume`
+- Secret containing the `ignition`
+
+## Prerequisites
+
+- [Butane](https://coreos.github.io/butane/)
+
+## Usage
+1. Adapt the `namespace` in `kustomization.yaml`
+2. Replace `your-user` [^1], `your-pw-hash`  [^2] and `your-ssh-key` [^3] in the `ignition/ignition.yaml`
```
afritzler, line 16: I would remove the footnotes as it should be clear what to change given the keys.

config/samples/e2e/machine-with-non-ephemeral-resource/README.md
```diff
@@ -0,0 +1,63 @@
...
+```
+patches:
+- path: patch-machineclassref.yaml
+- path: patch-volume.yaml
+```
+
+7. Run (`kubectl apply -k ./`) 
+
+
+[^1]: e.g. `max`
```
afritzler, line 56: Those footnotes can be removed.

internal/app/app_suite_test.go
```diff
@@ -32,7 +32,7 @@ import (
 
 const (
 	pollingInterval      = 50 * time.Millisecond
-	eventuallyTimeout    = 5 * time.Second
+	eventuallyTimeout    = 10 * time.Second
```
afritzler, line 35: Any reason why we should increase the timeout here?

afritzler (changes requested):

config/samples/e2e/machine-with-non-ephemeral-resource/README.md
Rohit-0505: done

config/samples/e2e/machine-with-non-ephemeral-resource/README.md
Rohit-0505: done

config/samples/e2e/machine-with-non-ephemeral-resource/README.md
Rohit-0505: done

internal/app/app_suite_test.go
Rohit-0505: There is one eventually block in `core_test.go` which throws timeout error with 5 seconds, so I thought to increase the timeout, also this error occurs sporadically with github runner. It does not give any timeout error while running tests in local machine.

Rohit-0505: Reverted the `eveltuallyTimeout` to 3s. Let's handle this timeout error in another PR and follow up in another issue.

afritzler (approved):

PR #1144, opened 2024-10-29
Add e2e example manifests for `Network` peering

ushabelgur:
# Proposed Changes

- Add e2e example manifests for `Network` peering

/ref https://github.com/ironcore-dev/ironcore/issues/1129

### DIFF
--- a/config/samples/e2e/bases/network-peering/machine1.yaml
+++ b/config/samples/e2e/bases/network-peering/machine1.yaml
@@ -1,0 +1,38 @@
+apiVersion: compute.ironcore.dev/v1alpha1
+kind: Machine
+metadata:
+  name: machine-sample1
+spec:
+  machineClassRef:
+    name: machineclass-sample
+  ignitionRef:
+    name: ignition
+    key: ignition.yaml
+  networkInterfaces:
+  - name: primary
+    ephemeral:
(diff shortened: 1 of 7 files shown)


config/samples/e2e/bases/network-peering/machine1.yaml
```diff
@@ -0,0 +1,38 @@
...
+          networkRef:
+            name: network-sample1
+          virtualIP:
+            ephemeral:
+              virtualIPTemplate:
+                spec:
+                  type: Public
+                  ipFamily: IPv4
+  volumes:
+  - name: root-disk-1
```
afritzler, line 29: Just call the disk `root-disk`.

config/samples/e2e/bases/network-peering/network2.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: networking.ironcore.dev/v1alpha1
+kind: Network
+metadata:
+  name: network-sample2
+  namespace: dev-team-onmetal-api
```
afritzler, line 5: Please use the same namespace here for the peering as the refs here also indicate that we are pairing within the same `namespace`.

afritzler (changes requested):

afritzler (approved):

PR #1145, opened 2024-10-29
Add e2e `IPAM` example manifests for `IPv4` and `IPv6`

Rohit-0505:
- Add IPAM example manifests for IP allocation to subnets of type IPv4
- Add IPAM example manifests for IP allocation to subnets of type IPv6

/ref #1129


config/samples/e2e/IPAM-example-IPv4/README.md
```diff
@@ -0,0 +1,10 @@
+# `IP` allocation to Subnets with `IPv4`
```
afritzler, line 1: Can we keep the directory names lower case.

config/samples/e2e/IPAM-example-IPv4/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
+namespace: test
```
afritzler, line 4: Please use the default `namespace` here (not specifying a namespace at all) as for this e2e it does not matter which namespace the resources are in.

afritzler (changes requested):

afritzler (changes requested):
Can we remove `example` from the folder name and just use `ipam-ipv4` and `ipam-ipv6`?

afritzler (approved):

PR #1146, opened 2024-10-29
Do not use process substitution in `hack/update-codegen.sh`

lukasfrank:
# Proposed Changes

- Fix codegen on macOS without `GNU bash`


hack/update-codegen.sh
```diff
@@ -81,11 +81,18 @@ for GV in ${ALL_VERSION_GROUPS}; do
   applyconfigurationgen_external_apis+=("${THIS_PKG}/api/${G}/${V}:${THIS_PKG}/client-go/applyconfigurations/${G}/${V}")
 done
 applyconfigurationgen_external_apis_csv=$(IFS=,; echo "${applyconfigurationgen_external_apis[*]}")
+
```
afritzler, line 84: Remove unnecessary newline.

afritzler (changes requested):

afritzler:
I would categorize it as a feature :)

afritzler (approved):

PR #1153, opened 2024-11-14
Introducing a `ReclaimPolicy` for ephemerally created `Volume` resource

### LINKED ISSUE #1114: Introduce a reclaim policy for ephemerally created `Volume` resources
# Summary

When creating a `Machine` resource there is the possibility to create the resource with an ephemerally managed `Volume` resource. The resulting `Volume` will show an `ownerReference` entry pointing to the created `Machine` resource, as well as `blockOwnerDeletion` set to `true`.

In case the `Machine` has to be recreated, there is currently no possibility to prevent deletion of the ephemerally managed `Volume` resource. This shortcoming should be tackled with the introduction of a `Reclaim Policy` similar to [Kubernetes Reclaim Policies for Persistent Volumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#reclaiming).

The `Reclaim Policy` should support 2 modes:
- `Retain`: the resource is not deleted after the managing resource has been deleted
- `Delete`: the current behavior, the resource is garbage-collected when the managing resource has been deleted

To not break current behavior the `Reclaim Policy` should be defaulted to `Retain` if not set.

## Basic example

```
apiVersion: compute.ironcore.dev/v1alpha1
kind: Machine
metadata:
  name: machine
spec:
  volumes:
  - device: oda
    ephemeral:
      volumeTemplate:
        metadata:
          creationTimestamp: null
        spec:
          reclaimPolicy: Retain # this is the new field defining the "Retain" strategy for the corresponding volume resource
          image: my-image:tag
          resources:
            storage: 10Gi
          volumeClassRef:
            name: fast
          volumePoolRef:
            name: pool
    name: root-disk-1
```

This instructs the controller to not remove the `Volume` resource. When the managing `Machine` resource is deleted, the `OwnerReference` as well as the `.spec.claimRef` should be cleared.

## Open questions

- should the controller label the `Volume` resource after deletion of the managing resource as such it is no longer `ephemerally managed`? 

## Motivation

This proposed feature would make it possible to safeguard for potential data loss in case the owning `Machine` resource needs to be recreated, independent of the reason (failure, config change, ...).

ushabelgur:
# Proposed Changes

- Add `ReclaimPolicy` for ephemerally created `Volume` resource with 2 supported modes: 
  - `Retain`: the resource is not deleted after the managing resource has been deleted 
  - `Delete`: the current behavior, the resource is garbage-collected when the managing resource has been deleted
- By default `ReclaimPolicy` would be assumed to be `Delete` to not to not break current behavior.
- Add wrapper type for `VolumeTemplateSpec`  to support `ReclaimPolicy` type along with existing VolumeSpec
- `OwnerReference` is added to ephemerally created volume in both the case of `ReclaimPolicy` is set to `Delete` to avoid deletion of ephemeral volumes accidently.
- Machine controller will take care of removing ownerRef from ephemerally created volumes if `ReclaimPolicy` is set to `Retain`.
- Add Test cases
[Note: volume_release_controller is already taking care of releasing volumes whose claimer doesn't existing, by setting .spec.claimRef to nil when claimer `Machine`  object is deleted. So not adding any extra logic for this]

Fixes #1114


internal/apis/storage/volume_types.go
```diff
@@ -122,8 +122,24 @@ type VolumeList struct {
...
+	VolumeSpec
+	// ReclaimPolicy defines the strategy for the corresponding volume resource after its managing resource has been deleted
+	ReclaimPolicy ReclaimPolicy
+}
+
 // VolumeTemplateSpec is the specification of a Volume template.
 type VolumeTemplateSpec struct {
 	metav1.ObjectMeta
-	Spec VolumeSpec
+	Spec EphemeralVolumeSpec
```
balpert89, line 144: Minor: please move this upwards in between `type VolumeAccess struct` and `type VolumeStatus struct` at line https://github.com/ironcore-dev/ironcore/blob/987deb12cd36ba28f1cbaa9da821df2885e70db3/api/storage/v1alpha1/volume_types.go#L58

internal/controllers/compute/machine_ephemeralvolume_controller.go
```diff
@@ -71,19 +72,29 @@ func (r *MachineEphemeralVolumeReconciler) ephemeralMachineVolumeByName(machine
...
 				Annotations: maps.Clone(ephemeral.VolumeTemplate.Annotations),
 			},
-			Spec: ephemeral.VolumeTemplate.Spec,
+			Spec: ephemeral.VolumeTemplate.Spec.VolumeSpec,
 		}
 		annotations.SetDefaultEphemeralManagedBy(volume)
-		_ = ctrl.SetControllerReference(machine, volume, r.Scheme())
+		if machineVolume.Ephemeral.VolumeTemplate.Spec.ReclaimPolicy != storagev1alpha1.Retain {
+			_ = ctrl.SetControllerReference(machine, volume, r.Scheme())
+		}
```
balpert89, line 80: if we only set the controller reference in case its `Reclaim` we are loosing the association for reconciliation. I am unsure if we can utilize https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/#set-orphan-deletion-policy because this is set during deletion of the controlling object, in this case when the `Machine` is deleted and has to be set by the caller.

We still have the `.spec.ClaimRef` as an association between `Machine` and `Volume` that can be sufficient for reconciliation means. Lets discuss in the sync call how to handle this.

utils/annotations/annotations.go
```diff
@@ -36,6 +36,10 @@ func IsDefaultEphemeralControlledBy(o metav1.Object, owner metav1.Object) bool {
 	return metav1.IsControlledBy(o, owner) && IsEphemeralManagedBy(o, commonv1alpha1.DefaultEphemeralManager)
 }
 
+func IsDefaultEphemeralOrControlledBy(o metav1.Object, owner metav1.Object) bool {
+	return metav1.IsControlledBy(o, owner) || IsEphemeralManagedBy(o, commonv1alpha1.DefaultEphemeralManager)
```
ushabelgur, line 40: Note: I added new method with OR condition here, as above original method IsDefaultEphemeralControlledBy() is used by all other ephemeral controllers and it will break behavior of those controllers.

internal/controllers/compute/machine_ephemeralvolume_controller.go
ushabelgur: As discussed when we set ownerRef, machine deletion also trigger setting the deletion timestamp in the owned objects, so we will handle it without adding ownerRef if reclaimPolicy is set to Retain.

internal/controllers/compute/machine_ephemeralvolume_controller.go
```diff
@@ -71,19 +72,29 @@ func (r *MachineEphemeralVolumeReconciler) ephemeralMachineVolumeByName(machine
 				Labels:      ephemeral.VolumeTemplate.Labels,
 				Annotations: maps.Clone(ephemeral.VolumeTemplate.Annotations),
 			},
-			Spec: ephemeral.VolumeTemplate.Spec,
+			Spec: ephemeral.VolumeTemplate.Spec.VolumeSpec,
 		}
 		annotations.SetDefaultEphemeralManagedBy(volume)
-		_ = ctrl.SetControllerReference(machine, volume, r.Scheme())
+		if machineVolume.Ephemeral.VolumeTemplate.Spec.ReclaimPolicy != storagev1alpha1.Retain {
```
afritzler, line 78: In case we use `Retain` policy for ephemeral `Volumes` and we don't set the controller reference, there will be no finalizer preventing the deletion of a Volume. We should prohibit the deletion of an ephemeral volume at all times as  long as the `Machine` object exists. One idea would be to always set the controller reference and remove it in the case of a `Machine` deletion.

afritzler (changes requested):

internal/controllers/compute/machine_ephemeralvolume_controller.go
lukasfrank: We might can utilize `RESTDeleteStrategy` in  `/k8s.io/apiserver@v0.30.7/pkg/registry/rest/delete.go`

ushabelgur: As discussed offline, If we want to handle removal of OwnerRef in registry using RESTDeleteStrategy, we are not adding ReclaimPolicy directly under VolumeSpec we have created wrapper around VolumeTemplateSpec as we didn't want to set this field for volumes other than ephemeral. so will not have reference to ReclaimPolicy in volme strategy. 
So handling removal of controller reference in machine controller before deleting machine.

ushabelgur:
As discussed offline, in ideal scenarios rook disks won't be created as ephemeral resource. In most of the case PVs will attached as volume to machine, which will be retained by default. This feature is not needed as of now. Closing the PR for now.

PR #1158, opened 2024-11-18
Add `requeueAfter` to volume deletion flow

### LINKED ISSUE #1155: Hotfix - add reconcile delay for volume processing
Processing volumes at scale requires a delay for the volumepoollet when an error occurs or another circumstance forces a requeue. See https://github.com/onmetal/onmetal-api/commit/29ddf38415e437e33340639d91d2127c7d43248d as a current hotfix

balpert89:
# Proposed Changes

- add a delay of 5 seconds when IRI volumes are not yet completely removed

Fixes #1155

### DIFF
--- a/poollet/volumepoollet/controllers/volume_controller.go
+++ b/poollet/volumepoollet/controllers/volume_controller.go
@@ -8,22 +8,24 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"time"
 
 	"github.com/go-logr/logr"
-	"github.com/ironcore-dev/controller-utils/clientutils"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
 	corev1alpha1 "github.com/ironcore-dev/ironcore/api/core/v1alpha1"
 	storagev1alpha1 "github.com/ironcore-dev/ironcore/api/storage/v1alpha1"
 	irimeta "github.com/ironcore-dev/ironcore/iri/apis/meta/v1alpha1"
 	iriVolume "github.com/ironcore-dev/ironcore/iri/apis/volume"
 	iri "github.com/ironcore-dev/ironcore/iri/apis/volume/v1alpha1"
-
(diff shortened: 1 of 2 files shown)


afritzler (changes requested):

poollet/volumepoollet/controllers/volume_controller.go
```diff
@@ -125,7 +127,7 @@ func (r *VolumeReconciler) deleteGone(ctx context.Context, log logr.Logger, volu
 	}
 	if !ok {
 		log.V(1).Info("Not all iri volumes are gone, requeueing")
-		return ctrl.Result{Requeue: true}, nil
+		return ctrl.Result{Requeue: true, RequeueAfter: 5 * time.Second}, nil
```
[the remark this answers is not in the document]
afritzler: Also 
```	
// RequeueAfter if greater than 0, tells the Controller to requeue the reconcile key after the Duration.
// Implies that Requeue is true, there is no need to set Requeue to true at the same time as RequeueAfter.
RequeueAfter time.Duration
```

poollet/volumepoollet/controllers/volume_controller.go
```diff
@@ -198,7 +200,7 @@ func (r *VolumeReconciler) delete(ctx context.Context, log logr.Logger, volume *
...
```
lukasfrank, line 203: tbh. I'm not sure if we should use a "fixed" delay, since we are working actively against the `controller-runtime` requeue mechanism with an exponential backoff. Waiting a fixed time multiplies if we build multiple layers of the `brokers/poollets`.  Wdyt about returning an error to circumvent this?

lukasfrank: I think even without returning an error we already have the optimal behavior: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/reconcile/reconcile.go#L99-L112

lukasfrank: // If the error is nil and result.RequeueAfter is zero and result.Requeue is true, the request
	// will be requeued using exponential backoff.

balpert89:
Closing, see #1155

PR #1163, opened 2024-11-19
Fix flapping IPs in NetworkInterface status

### LINKED ISSUE #1111: Flapping IPs in `NetworkInterface` status
**Describe the bug**

We are currently experiencing a flapping behavior of the `NetworkInterface` status. Once an `IP` is assigned in the status of the object it sometimes keeps changing from `<None>` to a valid IP - and vice versa. The problem seems be be occurring in a brokered setup, meaning we have a `machinepoollet/machinebroker` between the `NetworkInterface` resource and the actual usage in e.g. in a `provider` context.

**Expected behavior**
Once an IP has been assigned to a `NetworkInterface` it should be stable until the `NetworkInterface` IP changes or is actively being removed.

A potential root cause could be the `poollet/broker` interaction. We could try to reproduce this issue in our unit test setup by using `Consistently` instead of `Eventually` matcher.

ushabelgur:
# Proposed Changes

- Update `NetworkInterface` status only when there is change in `State` or list of `IPs`

Fixes #1111

### DIFF
--- a/poollet/machinepoollet/controllers/controllers_suite_test.go
+++ b/poollet/machinepoollet/controllers/controllers_suite_test.go
@@ -54,7 +54,7 @@ var (
 const (
 	eventuallyTimeout    = 3 * time.Second
 	pollingInterval      = 50 * time.Millisecond
-	consistentlyDuration = 1 * time.Second
+	consistentlyDuration = 10 * time.Second
 	apiServiceTimeout    = 5 * time.Minute
 
 	controllerManagerService = "controller-manager"
--- a/poollet/machinepoollet/controllers/controllers_suite_test.go
+++ b/poollet/machinepoollet/controllers/controllers_suite_test.go
@@ -54,7 +54,7 @@ var (
 const (
 	eventuallyTimeout    = 3 * time.Second
 	pollingInterval      = 50 * time.Millisecond
-	consistentlyDuration = 10 * time.Second
+	consistentlyDuration = 3 * time.Second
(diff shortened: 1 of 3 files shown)


balpert89:
suggestion: the consistently threshold is set to 1 second (see https://github.com/ironcore-dev/ironcore/blob/35d01e526b25a60fe6a72125e3ad7b3c52363416/poollet/machinepoollet/controllers/controllers_suite_test.go#L57). Please increase this to 5-10 seconds as the flapping could occur after a few seconds.

poollet/machinepoollet/controllers/controllers_suite_test.go
```diff
@@ -54,7 +54,7 @@ var (
 const (
 	eventuallyTimeout    = 3 * time.Second
 	pollingInterval      = 50 * time.Millisecond
-	consistentlyDuration = 1 * time.Second
+	consistentlyDuration = 10 * time.Second
```
afritzler, line 57: 10 seconds is quite a long time IMO. As the polling interval is currently set to 50 milliseconds maybe we should stick with 3 seconds (same as the eventually timeout).

afritzler (changes requested):

poollet/machinepoollet/controllers/controllers_suite_test.go
ushabelgur: done

afritzler (approved):

PR #1169, opened 2024-11-22
Fix IRI `*-broker` status

so-sahu:
# Proposed Changes

This PR addresses an issue in the getTargetIronCoreMachinePools method within the broker status IRI response. Currently, the method retrieves and returns a list of all MachinePools in the cluster, even when a specific MachinePool is provided. This behavior occurs due to a missing return statement when a single MachinePool is fetched.


balpert89:
LGTM

lukasfrank (changes requested):
Thanks for the fix, LGTM. Do you mind fixing it for the other brokers as well in this PR?

lukasfrank (approved):

afritzler (approved):

PR #1206, opened 2024-12-19
Fix missing `admissionregistration` RBAC

lukasfrank:
# Proposed Changes

- Validating Admission Policy turned from beta to stable in version 1.30 and apiserver needs access to the resources.


lukasfrank:
duplicate of #1205

PR #1215, opened 2025-01-09
Add usage doc for `Network` and `LoadBalancer`

ushabelgur:
# Proposed Changes

- Add usage documentation for `Network` and `LoadBalancer` in network section.
- Sample example yaml to create each of the resources with keyFields explained
- Reconciliation logic explained for each resource.

/ref #1191

### DIFF
--- a/docs/usage/networking/loadbalancer.md
+++ b/docs/usage/networking/loadbalancer.md
@@ -20,9 +20,6 @@ spec:
       app: web
   ports:
   - port: 80
-#status:
-#  ips:
-#  - 10.0.0.1 # The publicly available IP of the load balancer
 
 ```
 (`Note`: Refer to <a href="https://github.com/ironcore-dev/ironcore/tree/main/config/samples/e2e/loadbalancer-public">E2E Examples</a> for more detailed examples.)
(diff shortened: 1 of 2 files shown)


docs/usage/networking/loadbalancer.md
```diff
@@ -1 +1,66 @@
...
+  type: Public
+  ipFamilies: [IPv4]
+  networkRef:
+    name: network-sample
+  networkInterfaceSelector:
+    matchLabels:
+      app: web
+  ports:
+  - port: 80
+#status:
```
afritzler, line 23: Can we remove the commented status here and rather state in a section after the example what goes into the status?

docs/usage/networking/network.md
```diff
@@ -1 +1,39 @@
...
+    networkRef:
+      name: network-sample2
+```
+
+# Key Fields:
+- `providerID`(`string`): providerID is the provider-internal ID of the network.
+- `peerings`(`list`): peerings are the list of network peerings with this network(Optional).
+- `incomingPeerings`(`list`): incomingPeerings is a list of PeeringClaimRefs which is nothing but peering claim references of other networks.
+
+# Reconciliation Process:
```
afritzler, line 27: Can we add here a few words how and by whom the `providerID` and handle is assigned?

afritzler (changes requested):

docs/usage/networking/loadbalancer.md
ushabelgur: done

docs/usage/networking/network.md
ushabelgur: Added more description for `providerID`

afritzler (approved):

PR #1217, opened 2025-01-13
Add validation for `Machine` network interface spec

### LINKED ISSUE #1186: Validate `PrefixLength` for `IPSource` to ensure a length of 32/128
Currently we allow IP sources length other than 32 (IPv4) or 128 (IPv6) when creating a machine with a derived prefix. This fails silently and prevents the machine NIC creation.

ushabelgur:
# Proposed Changes

- Add validation for machine spec network interface
(Note: `NetworkInterface`  already has the validation for IP sources length i.e. only single IP is allowed https://github.com/ironcore-dev/ironcore/blob/6bbca962fea8e92b279304acae43c6f67ff2b323/internal/apis/networking/validation/networkinterface.go#L143
added missing validation for machine spec networkInterface which will cover prefixLength validation as well)

Fixes #1186

### DIFF
--- a/internal/apis/compute/validation/machine.go
+++ b/internal/apis/compute/validation/machine.go
@@ -9,6 +9,8 @@ import (
 	"github.com/ironcore-dev/ironcore/internal/admission/plugin/machinevolumedevices/device"
 	ironcorevalidation "github.com/ironcore-dev/ironcore/internal/api/validation"
 	"github.com/ironcore-dev/ironcore/internal/apis/compute"
+	"github.com/ironcore-dev/ironcore/internal/apis/networking"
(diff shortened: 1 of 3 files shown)


balpert89:
LGTM

afritzler (changes requested):
Can we add in the test cases an `ephemeral` NIC which derives it's IP from a parent prefix with a length of 32/128 (this will be valid) and values which are greater than this and those will be invalid.

ushabelgur:
> Can we add in the test cases an `ephemeral` NIC which derives it's IP from a parent prefix with a length of 32/128 (this will be valid) and values which are greater than this and those will be invalid.

Added test cases

afritzler (approved):

PR #1222, opened 2025-01-21
Introduce `ReclaimPolicy` for ephemeral `VirtualIP`

### LINKED ISSUE #1115: Introduce a reclaim policy for ephemerally created `Virtual IP` resources
# Summary

Similar to https://github.com/ironcore-dev/ironcore/issues/1114 the same `Reclaim Policy` should be available for `Virtual IP` resources.

## Basic example

```
apiVersion: compute.ironcore.dev/v1alpha1
kind: Machine
metadata:
  name: machine
spec:
  networkInterfaces:
  - ephemeral:
      networkInterfaceTemplate:
        metadata:
          creationTimestamp: null
        spec:
          ipFamilies:
          - IPv4
          ips:
          - ephemeral:
              prefixTemplate:
                metadata:
                  creationTimestamp: null
                spec:
                  ipFamily: IPv4
                  parentRef:
                    name: prefix
                  prefixLength: 32
          networkRef:
            name: network
          virtualIP:
            reclaimPolicy: Retain # this is the new field defining the "Retain" strategy for the corresponding volume resource
            virtualIPRef:
              name: ip-ref
    name: primary
```

This instructs the controller to not remove the `Virtual IP` resource. When the managing `Machine` / `NetworkInterface` resource is deleted, the OwnerReference as well as the `.spec.targetRef` should be cleared.

## Motivation

Resource saving for consuming teams, such as VPN configuration.

Rohit-0505:
# Proposed Changes

- Add `ReclaimPolicy` for ephemerally created `virtualIP` resource with below supported modes:
  - Retain: the resource is not deleted after the managing resource has been deleted
  - Delete: the current behavior, the resource is garbage-collected when the managing resource has been deleted
- The default `ReclaimPolicy` is considered implicit `Delete` to avoid breaking current behavior.
- Add wrapper type for `VirtualIPTemplateSpec` to support `ReclaimPolicy` type along with existing VirtualIPSpec
- OwnerReference is set to an ephemerally created virtualIP with ReclaimPolicyTypeDelete.
- To prevent deletion of ephemeral vip without ownerReference a RESTDeleteStrategy method ValidateDelete is added in the strategy.go
- Add Test cases
[Note: `virtualIP_release_controller` is already taking care of releasing virtualIP whose claimer doesn't exist, by setting .spec.targetRef to nil when claimer NetworkInterface object is deleted. So not adding any extra logic for this]

Fixes #1115


api/networking/v1alpha1/virtualip_types.go
```diff
@@ -55,8 +65,15 @@ type VirtualIPList struct {
 	Items           []VirtualIP `json:"items"`
 }
 
+type EphemeralVirtualIPSpec struct {
+	// VirtualIPSpec defines the desired state of VirtualIP
```
afritzler, line 69:
```suggestion
	// VirtualIPSpec defines the desired state of a VirtualIP
```

api/networking/v1alpha1/virtualip_types.go
```diff
@@ -55,8 +65,15 @@ type VirtualIPList struct {
...
+	VirtualIPSpec `json:",inline"`
+	// ReclaimPolicy is the ReclaimPolicyType of virtualIP
```
afritzler, line 71:
```suggestion
	// ReclaimPolicy is the ReclaimPolicyType of a VirtualIP
```

internal/apis/networking/virtualip_types.go
```diff
@@ -55,8 +65,15 @@ type VirtualIPList struct {
 	Items []VirtualIP
 }
 
+type EphemeralVirtualIPSpec struct {
+	// VirtualIPSpec defines the desired state of VirtualIP
```
afritzler, line 69:
```suggestion
	// VirtualIPSpec defines the desired state of a VirtualIP
```

internal/apis/networking/virtualip_types.go
```diff
@@ -55,8 +65,15 @@ type VirtualIPList struct {
...
+	VirtualIPSpec
+	// ReclaimPolicy is the ReclaimPolicyType of virtualIP
```
afritzler, line 71:
```suggestion
	// ReclaimPolicy is the ReclaimPolicyType of a VirtualIP
```

api/networking/v1alpha1/virtualip_types.go
```diff
@@ -55,8 +65,15 @@ type VirtualIPList struct {
...
+	ReclaimPolicy ReclaimPolicyType `json:"reclaimPolicy,omitempty"`
```
afritzler, line 72: What should happen if I don't provider a Reclaim policy? Do we default to `Delete`? Or do we implicitly use the `Delete` behaviour?

internal/controllers/networking/networkinterface_ephemeralvirtualip_controller_test.go
```diff
@@ -66,6 +70,143 @@ var _ = Describe("NetworkInterfaceEphemeralVirtualIP", func() {
...
+		Expect(k8sClient.Delete(ctx, nic)).To(Succeed())
+
+		By("ensuring the nic is deleted")
+		nicKey := client.ObjectKey{Namespace: ns.Name, Name: nic.Name}
+		err := k8sClient.Get(ctx, nicKey, nic)
+		Expect(apierrors.IsNotFound(err)).To(BeTrue())
+
+		By("ensuring the virtualIP still exists after the nic is deleted")
+		vipKey := client.ObjectKey{Namespace: ns.Name, Name: vip.Name}
+		Expect(k8sClient.Get(ctx, vipKey, vip)).To(Succeed())
```
afritzler, line 208: We should probably add here a test case for the case that the `RetainPolicy` is not set.

afritzler (changes requested):

api/networking/v1alpha1/virtualip_types.go
Rohit-0505: done

api/networking/v1alpha1/virtualip_types.go
Rohit-0505: done

internal/apis/networking/virtualip_types.go
Rohit-0505: done

internal/apis/networking/virtualip_types.go
Rohit-0505: done

api/networking/v1alpha1/virtualip_types.go
Rohit-0505: We don't set the default `ReclaimPolicy` explicitly to `Delete`, It's considered implicit `Delete` to avoid breaking current behavior.

internal/controllers/networking/networkinterface_ephemeralvirtualip_controller_test.go
Rohit-0505: This test case is already present if the `ReclaimPolicy` is not set, which means it's an implicit `Delete`. We then validate the ownerRef, etc.
Also for a vip having `ReclaimPolicy` set to Retain, if later updated with `ReclaimPolicyDelete`, a test case is added to validate it.

afritzler (approved):

PR #1329, opened 2025-07-16
Fix collision in `Bucket` secret name

### LINKED ISSUE #1327: Bucket secret name conflict is created after having onmetal and ironcore type buckets in single namespace
**Describe the bug**
A conflict is created for bucket secret name after having onmetal and ironcore type buckets in single namespace.

**To Reproduce**
1. Create onmetal bucket in namespace
2. Create ironcore bucket in same namespace
3. Ironcore bucket will remain in pending state because of secret name conflict

**Expected behavior**
Onmetal and Ironcore bucket creation should be successful

kasabe28:
# Proposed Changes

- change bucket secret name to bucket uid

Fixes https://github.com/ironcore-dev/ironcore/issues/1327

### DIFF
--- a/poollet/bucketpoollet/controllers/bucket_controller.go
+++ b/poollet/bucketpoollet/controllers/bucket_controller.go
@@ -5,8 +5,6 @@ package controllers
 
 import (
 	"context"
-	"crypto/sha256"
-	"encoding/hex"
 	"fmt"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


poollet/bucketpoollet/controllers/bucket_controller.go
```diff
@@ -388,11 +386,6 @@ func (r *BucketReconciler) create(ctx context.Context, log logr.Logger, bucket *
 	return ctrl.Result{}, nil
 }
 
-func (r *BucketReconciler) bucketSecretName(bucketName string) string {
-	sum := sha256.Sum256([]byte(bucketName))
-	return hex.EncodeToString(sum[:])[:63]
-}
-
```
lukasfrank, line 395: What about doing it the same way like in the `Volume` reconciler? 
```
func (r *VolumeReconciler) volumeSecretName(volumeName, volumeHandle string) string {
	sum := sha256.Sum256([]byte(fmt.Sprintf("%s/%s", volumeName, volumeHandle)))
	return hex.EncodeToString(sum[:])[:63]
}
```

lukasfrank (approved):

PR #1330, opened 2025-07-16
Fix sporadic test failures in `machinepoollet` test suite

### LINKED ISSUE #1323: Flaky tests
**Describe the bug**
The `IronCore` tests seem to be flaky which should be addressed: 

Ref:
https://github.com/ironcore-dev/ironcore/actions/runs/16135575258/job/45531082871?pr=1317#step:4:592

ushabelgur:
# Proposed Changes

- As common `MachinePool` is being used across all controller package testcases, sometime MachineClasses attached to the pool are not yet deleted and we are doing equal comparison of list of resources, which is causing test case to fail sometimes.
- Instead of equal added contains check to see if MachinePool contains  newly created MachineClasses with expected quantity 

Fixes #1323

### DIFF
--- a/poollet/machinepoollet/controllers/machine_controller_test.go
+++ b/poollet/machinepoollet/controllers/machine_controller_test.go
@@ -44,6 +44,7 @@ var _ = Describe("MachineController", func() {
 			},
 		}
 		Expect(k8sClient.Create(ctx, network)).To(Succeed())
+		DeferCleanup(k8sClient.Delete, network)
 
 		By("patching the network to be available")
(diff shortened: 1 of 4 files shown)


ushabelgur:
I see some other tests failing after fixing machinePoollet related tests 
Like this https://github.com/ironcore-dev/ironcore/actions/runs/16320623702/job/46097150773?pr=1330 and 
this https://github.com/ironcore-dev/ironcore/actions/runs/16335652712/job/46147139835?pr=1330
Should fix them as well

afritzler (approved):

PR #1363, opened 2025-08-19
Remove `downward-api` label filtering dependency

### LINKED ISSUE #1339: Remove dependency to `downward-api.machinepoollet.ironcore.dev` for filtering
# Summary

As in this discussion ( https://github.com/ironcore-dev/ironcore/pull/1335#discussion_r2222704412 ) pointed out, relying on the `downward-api.machinepoollet.ironcore.dev` is not optimal.

The `downward-api` is only intended for ops purposes and can be configured to have a different name or it might be possible that the labels are not present at all (see: `machine-downward-api-label`). 

## Proposed change

At the moment we only store the _parent_ `Machine.UID` in `metadata.annotations.machinebroker.ironcore.dev/labels`. We can additionally put the `machinepoollet.ironcore.dev/machine-uid` as a label to the `Machine` and utilize this for filtering. 

## Tasks

- [ ] Validate if this changes can work

**Poollet:**
- [ ] Update  `listMachinesByMachineKey` to use again a `machineUIDLabelSelector`

**Broker:**
- [ ] Simplify `ListMachines` 
  - [ ] Remove this check `labelFilter := req.Filter; labelFilter != nil && labelFilter.LabelSelector != nil`  and handle it unified
- [ ] Update `CreateMachine` to set the label to the `Machine`

kasabe28:
# Proposed Changes

- Modify `machinepoollet` to list and delete machines using `machinepoollet.ironcore.dev/machine-uid` label
- Simplify `ListMachines` by handling it unified
- Modify `machinebroker` to add `machinepoollet.ironcore.dev/machine-uid` label with parent machine uid as value in `CreateMachine` function
- Update tests 
- Change default iriGeneration from `1` to `0`

Fixes #1339

### DIFF
--- a/poollet/machinepoollet/controllers/machine_controller.go
+++ b/poollet/machinepoollet/controllers/machine_controller.go
@@ -34,7 +34,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/kubectl/pkg/util/fieldpath"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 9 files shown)


poollet/machinepoollet/controllers/machine_controller.go
```diff
@@ -93,37 +86,26 @@ func (r *MachineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-	return res.Machines, nil
-}
-
-func (r *MachineReconciler) listMachinesByMachineKey(ctx context.Context, machineKey client.ObjectKey) ([]*iri.Machine, error) {
-	res, err := r.MachineRuntime.ListMachines(ctx, &iri.ListMachinesRequest{
-		Filter: &iri.MachineFilter{LabelSelector: r.machineKeyLabelSelector(machineKey)},
-	})
-	if err != nil {
-		return nil, fmt.Errorf("error listing machines by machine key: %w", err)
+	if machine.Status.MachineID != "" {
```
lukasfrank, line 95: This check is kind of bad since in case of a controller crash we lose the connection between two `Machine`s.  We can not get over time in a sable state again. The scenario is not that unlikely, let's assume you create a new `Machine`
- `getIRIMachinesForMachine` has len 0 
- `func (r *MachineReconciler) create(` will triggered 
- `Machine` in the below cluster will be created
-  But something within fails: 
```
	log.V(1).Info("Updating status")
	if err := r.updateStatus(ctx, log, machine, res.Machine, nics); err != nil {
		return ctrl.Result{}, fmt.Errorf("error updating machine status: %w", err)
	}
```
In the next reconcile we again recreate the `Machine` which leads to a unstable state

lukasfrank (changes requested):

lukasfrank (approved):

Issue #1343, opened 2025-07-24
Integrate `conditions` API for all relevant API resources in IronCore


balpert89:
# Summary

The [meta package in K8S apimachinery](https://pkg.go.dev/k8s.io/apimachinery@v0.33.3/pkg/apis/meta/v1#Condition) exposes a `Condition` struct which enriches status fields with information such as `transitionTime`, different fine-granular condition types and reasons why this condition has been set.

The [controller-utils package](https://github.com/ironcore-dev/controller-utils/tree/main/conditionutils) offers utility functions around conditions.

## Basic example

```
// either introduce our own condition type or leverage the apimachinery condition type
type ConditionWithTimestamps struct {
	Type               string
	Status             corev1.ConditionStatus
	LastUpdateTime     *metav1.Time
	LastTransitionTime *metav1.Time
	Reason             string
	Message            string
}

type MachineStatus struct {
	...
	Conditions []ConditionWithTimestamps
}
```

Naming is up for suggestions.

## Motivation

The IronCore API should start actively leveraging this feature to enhance operational observability of API resources.

balpert89:
as discussed, we won't do this, the added value is debatable. I will also close the associated PR.

PR #1424, opened 2025-12-01
Integrate `ConditionsAPI` to `volume` resource

Rohit-0505:
# Proposed Changes

- Update `volume_types` with ConditionsAPI to integrate with `VolumeStatus`
- Implement the conditions for volume in the volumepoollet
- Validate the changes with test cases
/ref #1343


Rohit-0505:
Closing this PR, adding conditions API for volume resource is not needed at the moment, since there are no dependent resources that would make use of these conditions in their controller reconciliation.

PR #1475, opened 2026-05-22
Add namespaces to different levels of deployable kustomizations

### LINKED ISSUE #1472: Implement Pool Health Proposal
# Summary

To determine the availability and health of each machine pool, we propose introducing heartbeats and status conditions of pools.

Similar to Kubernetes' node heartbeats & status conditions, there are 2 parties involved in this:

- The pool itself, regularly reporting its status by patching the *Pool resource's status and acquiring / keeping a lease object within the ironcore-machinepool-lease namespace. For references on leases, see the [leases Kubernetes documentation](https://kubernetes.io/docs/concepts/architecture/leases/).
 - A pool controller within the IronCore control plane that updates a pool object's status conditions depending on the heartbeats of the pool.

It was decided to only implement this mechanism for `MachinePool`s.

## ToDos

- [x] Implementation -> https://github.com/ironcore-dev/ironcore/pull/1476
  - [x] heartbeats
    - [x] machinepoollet creates and updates Lease in `ironcore-machinepool-lease` namespace
  - [x] controllers ironcore/controller/*pool
     - [x] machinepool-lifecycle-controller that monitors the Lease and Ready Condition
- [x] kustomize config adjustments
- [x] bump and adopt in ironcore-in-a-box -> https://github.com/ironcore-dev/ironcore-in-a-box/pull/93
- [x] document -> https://github.com/ironcore-dev/ironcore-dev.github.io/pull/137

## Motivation

See the [proposal's motivation section](https://github.com/ironcore-dev/enhancements/blob/main/ieps/15-pool-health.md#motivation)

friegger:
Deal with namespaces more explicitly by adding them to kustomizations on different levels directly, eliminating the previous remove-namespace.yaml patch dance.

`apiserver` and `controller` components now have:
- config/\<component\>/default/   - namespace+namePrefix only, emits no
                                  Namespace (used as a base)
- config/\<component\>/standalone/ - wraps default/ and adds the
                                  ironcore-system Namespace; use this for
                                  single-component deploys

Shared Namespace lives in config/namespaces/ironcore-system/. The combined config/default and config/etcdless installs reference the bases and the  Build output for both is byte-identical to main (only kustomize deprecation-warning lines disappear, since the patches that triggered them are gone).

The namespaces folder can be extended with additional namespaces, e.g. for the pool leases.

Behavioral change for downstream consumers of the low level kustomizations: users who previously ran `kustomize build config/controller/default` for a complete deploy must migrate to `config/controller/standalone`; same for `config/apiserver/default` -> `config/apiserver/standalone` (or `config/apiserver/standalone-etcdless` for the external-etcd variant).

Consumers must not apply namespace transformers in overlays, these will fail once we have multiple namespaces. This is true, independent of this change, once base kustomizations provide multiple namespaces.

The Makefile install/uninstall/deploy/undeploy targets are retargeted at the standalone variants accordingly. hack/validate-kustomize.sh is also made portable (GNU realpath --relative-to is unavailable on macOS).

Contributes to #1472 

## Summary by CodeRabbit

* **New Features**
  * Added standalone and controller-only deployment overlays and a dedicated system namespace so users can deploy apiserver-only or controller-only setups.

* **Chores**
  * Installer/deployer defaults updated to use standalone overlays.
  * Removed duplicate/implicit namespace declarations for cleaner installs.
  * Improved validation messages to show clearer path information.


config/namespaces/ironcore-system/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
```
gonzolino, line 1: Missing SPDX license header. All other new files in this PR include it — e.g. `config/namespaces/ironcore-system/namespace.yaml` has:
```
# SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and IronCore contributors
# SPDX-License-Identifier: Apache-2.0
```
Please add the same two lines at the top of this file for consistency.

friegger: The [add-license make target](https://github.com/ironcore-dev/ironcore/blob/caaf1d3eb49df102637235e9efcb3c3a374fd7f5/Makefile#L125) only considers `*.go` files. I wonder, should every file contain it instead?

gonzolino: OK, if we only consider go files we don't need to add it here. But then I would not add the header to other yaml files, too(e.g. `config/etcdless/kustomization.yaml`).

friegger: Done, removed.

Makefile
```diff
@@ -228,20 +228,20 @@ docker-push: ## Push docker image with the manager.
 .PHONY: install
 install: manifests kustomize ## Install API server & API services into the K8s cluster specified in ~/.kube/config. This requires APISERVER_IMG to be available for the cluster.
 	cd config/apiserver/server && $(KUSTOMIZE) edit set image apiserver=${APISERVER_IMG}
-	kubectl apply -k config/apiserver/default
+	kubectl apply -k config/apiserver/standalone
 
 .PHONY: uninstall
 uninstall: manifests kustomize ## Uninstall API server & API services from the K8s cluster specified in ~/.kube/config.
-	kubectl delete -k config/apiserver/default
+	kubectl delete -k config/apiserver/standalone
```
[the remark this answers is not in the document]
friegger: We accept that for now.

config/namespaces/ironcore-system/kustomization.yaml
gonzolino: Now it is removed everywhere except this file 😅

friegger: Ah, that file was still sitting unsaved in the IDE. 🙈 Really done now.

gonzolino (approved):

PR #1500, opened 2026-07-03
Fix empty action field on IRI-mapped events

### LINKED ISSUE #1489: Event Handling has to be updated after breaking change in controller-runtime
**Describe the bug**
Release v0.4.0 updated controller-runtime to v0.23.x which brings a breaking change:
- Events: Migration to the new events API by @clebs in https://github.com/kubernetes-sigs/controller-runtime/pull/3262
  - Using the new GetEventRecorderFor requires updating your rbac for events to use the events.k8s.io apiGroup rather than the `` (core) apiGroup

Currently we see errors like this from controllers in the machine poollet:
```
1 event_broadcaster.go:270] "Server rejected event (will not retry!)" err="events.events.k8s.io \"webapp.18bba95a36d521da\" is forbidden: User \"system:serviceaccount:libvirt-provider:libvirt-provider-controller-manager\" cannot patch resource \"events\" in API group \"events.k8s.io\" in the namespace \"default\"" event="&Event{...}"
```

The events.k8s.io apiGroup also make the `action` field required. Events where the field is not set are rejected:
```
E0629 11:14:07.923183       1 event_broadcaster.go:270] "Server rejected event (will not retry!)" err="Event \"webapp.18bba95a36d521da\" is invalid: action: Required value" event="&Event{...}"
```
This part requires changes in the code.

**To Reproduce**
Use ironcore-in-a-box and create a machine. Check the logs of the libvirt-provider pod.

**Expected behavior**
No such error logs.

**ToDos**
- [x] RBAC changes for controllers in ironcore and its manifests -> #1490 
- [x] config file in libvirt-provider (service account is also used by machinepoollet)
- [x] Ensure `action` field is set for all events
  - [x] libvirt-provider
  - [x] ceph-provider

friegger:
The events.k8s.io/v1 API server rejects events with an empty action field. Providers do not always populate Spec.Action on IRI events, causing the poollet event mappers to push invalid events. Fall back to Spec.Reason when Spec.Action is empty.

Additional changes will update provider-utils and providers to correctly send the action field.

Contributes to #1489

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured emitted Kubernetes events always have a non-empty `action` value.
  * When the source event action is empty, Kubernetes event `action` now defaults to `Unknown`.
  * Applied consistently to bucket, machine, and volume event reporting.
  * Added automated tests confirming Kubernetes `Reason`, `Message`, and `Type` are preserved while `Action` is set to `Unknown` when the source action is empty.

### DIFF
--- a/poollet/bucketpoollet/bem/bem.go
+++ b/poollet/bucketpoollet/bem/bem.go
@@ -53,12 +53,12 @@ func (m *BucketEventMapper) relist(ctx context.Context, log logr.Logger) error {
 					Namespace: bucketEvent.Spec.InvolvedObjectMeta.Labels[v1alpha1.BucketNamespaceLabel],
 				},
 			}
-			// events.k8s.io/v1 requires a non-empty action field. Providers do not always populate
-			// Spec.Action on IRI events, so fall back to the reason to keep events acceptable to the
-			// API server.
+			// events.k8s.io/v1 requires a non-empty action field. Providers should populate
+			// Spec.Action on IRI events; fall back to "Unknown" so a forgotten value is
+			// obvious in the emitted event rather than silently dropped by the API server.
 			action := bucketEvent.Spec.Action
 			if action == "" {
-				action = bucketEvent.Spec.Reason
+				action = "Unknown"
 			}
 			m.Eventf(involvedBucket, nil, bucketEvent.Spec.Type, bucketEvent.Spec.Reason, action, bucketEvent.Spec.Message)
 		}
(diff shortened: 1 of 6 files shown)


lukasfrank:
What should happen if `Action` and `Reason` are empty strings?

friegger:
> What should happen if `Action` and `Reason` are empty strings?

Good point, changed the implementation to fallback to `Unknown`, if the Action is missing.

lukasfrank (approved):

PR #1505, opened 2026-07-06
Fix aggregated API server readiness race causing sporadic 503s in envtest suites

afritzler:
# Proposed Changes

Fix aggregated API server readiness race causing sporadic 503s in envtest suites.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved aggregated API server readiness checks by running real `List` round-trip probes for each discovered group/version.
  * Updated readiness polling to require multiple consecutive successful probes before marking endpoints ready, reducing false positives.
  * Refined retry behavior to continue through transient API/transport failures, while stopping immediately on non-transient errors.

### DIFF
--- a/utils/envtest/envtest.go
+++ b/utils/envtest/envtest.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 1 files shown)


afritzler:
@coderabbitai review

utils/envtest/envtest.go
```diff
@@ -661,6 +667,110 @@ func WaitUntilAPIServicesReady(ctx context.Context, ext *EnvironmentExtensions,
...
+	for _, gv := range gvs {
+		list, ok := newListForGroupVersion(scheme, gv)
+		if !ok {
+			continue
+		}
+		lists = append(lists, list)
+	}
+	if len(lists) == 0 {
+		return nil
+	}
```
lukasfrank: Doesn't lead this to issues? Right now, we fetch once and use the data for all retries, if the data is stale it fails continuously

adracus (approved):

PR #1509, opened 2026-07-10
Schedule ephemeral volumes in the same zone/region as the machine

### LINKED ISSUE #1497: Ephemeral Volumes end up in a different zone than the Machine
**Describe the bug**

When creating a machine with an ephemeral volume template and the machine has a topology label to be located in zone a, the volume can end up in zone b as it does not carry over the topology label from the machine.

**To Reproduce**

In a multi-zone IronCore installation with per-zone volume pools, create a machine with an ephemeral volume without a topology label but add a topology label to the machine.

Depending on RNG and load the volume may end up in a different zone.

**Expected behavior**

The volume and machine are always in a zone and a ephemeral volume created from a machine automatically carries over the topology labels.

gonzolino:
# Proposed Changes

Propagate topology labels from machine to ephemeral volumes

Copy zone and region labels from `MachinePoolSelector` to `VolumePoolSelector` when creating ephemeral volumes, without overwriting labels already set on the volume.

This ensures ephemeral volumes are scheduled in the same zone/region as the machine they belong to, except if the user has specifically defined different topology labels.

Fixes #1497

## Summary by CodeRabbit

* **New Features**
  * Ephemeral volumes now inherit applicable zone and region settings from their machine pool configuration.
  * Existing volume-specific topology settings are preserved and take precedence over inherited values.

* **Bug Fixes**
  * Improved ephemeral volume placement to ensure topology preferences are applied consistently.

### DIFF
--- a/internal/controllers/compute/machine_ephemeralvolume_controller.go
+++ b/internal/controllers/compute/machine_ephemeralvolume_controller.go
@@ -33,6 +33,11 @@ const (
 	MachineEphemeralVolumeManager = "machine-ephemeral-volume"
 )
 
+var poolSelectorTopologyLabels = []string{
+	string(commonv1alpha1.TopologyLabelZone),
+	string(commonv1alpha1.TopologyLabelRegion),
+}
+
 type MachineEphemeralVolumeReconciler struct {
 	k8sevents.EventRecorder
 	client.Client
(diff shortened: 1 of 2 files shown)


maxmoehl (approved):
> This ensures ephemeral volumes are scheduled in the same zone/region as the machine they belong to, except if the user has specifically defined different topology labels.

An important caveat is that this does not cover the case where the user did not put any labels into the selector. In that case the volume and machine could still end up in separate zones, no?

Otherwise, this is a worthwhile improvement, thank you!

lukasfrank (approved):
LGTM

gonzolino:
> An important caveat is that this does not cover the case where the user did not put any labels into the selector. In that case the volume and machine could still end up in separate zones, no?

Yes good point, this fix will only apply for situations were topology labels are given. I am not sure how to solve it for cases were we don't have them 🤔

PR #1536, opened 2026-08-05
Enhance scheduling to respect `MachinePool` `Ready` condition

### LINKED ISSUE #1531: Don't schedule on an unhealthy MachinePool
# Summary

#1472 introduced a mechanism to mark MachinePools as not ready. Consider the Ready condition for scheduling decisions.

## Motivation

Prevent scheduling machines on not-ready MachinePools.

lukasfrank:
# Proposed Changes

- Enhance machine scheduling to respect `MachinePool` `Ready` condition
  - Skip pools which are not ready of if the condition is missing

Fixes #1531

## Summary by CodeRabbit

* **Bug Fixes**
  * Machines are now scheduled only to machine pools that report readiness.
  * Unready pools are excluded from scheduling until they become ready, even when capacity is available.
  * Scheduling resumes automatically once a pool reaches the ready state.

### DIFF
--- a/internal/controllers/compute/machine_scheduler_test.go
+++ b/internal/controllers/compute/machine_scheduler_test.go
@@ -594,7 +594,7 @@ var _ = Describe("MachineScheduler", func() {
 		))
 	})
 
-	It("should not schedule machines onto a machine pool that is not ready", func(ctx SpecContext) {
+	It("should not schedule machines onto a machine pool without ready condition", func(ctx SpecContext) {
 		By("creating a machine pool")
 		machinePool := &computev1alpha1.MachinePool{
 			ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


afritzler:
@coderabbitai review

internal/controllers/compute/machine_scheduler_test.go
```diff
@@ -566,4 +593,93 @@ var _ = Describe("MachineScheduler", func() {
 			HaveField("Spec.MachinePoolRef.Name", Equal(secondMachinePool.Name)),
 		))
 	})
+
+	It("should not schedule machines onto a machine pool that is not ready", func(ctx SpecContext) {
```
gonzolino: I think I would update the test description to have a clear separation from the "ready condition is false" test case.

```suggestion
	It("should not schedule machines onto a machine pool without ready condition", func(ctx SpecContext) {
```

gonzolino (commented):
lgtm, one minor nit regarding the test description after adding the second test case.

internal/controllers/compute/machine_scheduler_test.go
lukasfrank: fixed

gonzolino (approved):

